PR #9, opened 2018-11-07
Fixed Docker and installation scripts.

sukritkalra:
- Added a `docker pull erdos` command to README.
- Fixed the Dockerfile to now clone the public repository instead.
- Fixed `install_dependencies.sh`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,20 +18,12 @@ ENV SHELL /bin/bash
 
 SHELL ["/bin/bash", "-c"]
 
-# Get ROS Kinetic for Ubuntu 16.04
-RUN sudo sh -c 'echo "deb http://packages.ros.org/ros/ubuntu xenial main" > /etc/apt/sources.list.d/ros-latest.list'
-RUN sudo apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key 421C365BD9FF1F717815A3895523BAEEB01FA116
-RUN sudo apt-get update && sudo apt-get -y install ros-kinetic-ros-base
-
-# Source the ROS setup script at the init of the shell.
-RUN echo "source /opt/ros/kinetic/setup.bash" >> ~/.bashrc
-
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -27,8 +27,9 @@ RUN sudo apt-get update && sudo apt-get -y install ros-kinetic-ros-base
 RUN echo "source /opt/ros/kinetic/setup.bash" >> ~/.bashrc
 
 # Get the erdos directory.
-RUN mkdir -p /home/erdos/workspace/erdos
-ADD . /home/erdos/workspace/erdos
+RUN sudo apt-get -y install git
+RUN mkdir -p /home/erdos/workspace
+RUN cd /home/erdos/workspace && git clone https://github.com/erdos-project/erdos.git
 RUN echo "export PYTHONPATH=\$PYTHONPATH:/home/erdos/workspace/erdos" >> ~/.bashrc
 WORKDIR /home/erdos/workspace/erdos
```
ICGog: Can you please invoke install_dependencies.sh in this Dockerfile? At the moment we duplicate the logic both.

README.md
```diff
@@ -8,48 +8,18 @@ ERDOS is a platform for building self-driving cars and controlling robots.
...
-To run the pipeline, do the following:
-
-```
-$ cd ~/workspace/CARLA_0_8_4/ && ./CarlaUE4.sh -windowed -ResX=8 -ResY=8 -carla-server
-$ cd ~/workspace/ERDOS/examples/benchmarks/pylot && python pylot.py --replay=True
-```
-
 
-# Local instalation
+# Local installation
```
ICGog: I think we should now be able to run on OS X when using Ray as back-end. Can you please update the readme, and ensure that the there's two installation options: with ROS and without ROS?

Dockerfile
sukritkalra: The Dockerfile now invokes install_dependencies.sh.

PR #10, opened 2018-11-07
Added several pre-defined operators: Concat, Map, Where, Unzip.

ICGog:


tests/operators_test.py
```diff
@@ -0,0 +1,114 @@
+import os
```
sukritkalra: Looks like `os` and `sys` are not being used.

Issue #7, opened 2018-11-06
Bug when stream names clash across operators


ICGog:
If multiple operators publish on data streams with the same name, then streams leak data from each other. 
The bug occurs because we do not anonymize stream names. When we use ROS as a back-end, we use the non-anonymized name to denote a ROS topic, and thus many operators can publish on the same topic. When we use Ray as a back-end, we use the non-anonymized `stream.name` to build several auxiliary data structures, and as a result we clash stream.

This behavior can be reproduces with `tests/stream_name_clash.py`.

PR #12, opened 2018-11-08
[Fix Bug] Stream name clashing problem

yujialuo:
Addressed https://github.com/erdos-project/erdos/issues/7
1. Rename duplicated stream names
2. change self.get_output_stream().send() into self.send()

Have not updated all examples using 2


erdos/op.py
```diff
@@ -50,6 +46,9 @@ def execute(self):
         """
         pass
 
+    def send(self, stream_name, msg):
```
ICGog: Can you please do now make this change in this PR? This change is unrelated to the bug, and makes the PR much larger. Plus, the examples are not going to work as a result of this change. I think it's better to separate this change into a new PR, in which we also update the examples. However, I would put this new PR as low priority for now.

erdos/graph.py
```diff
@@ -189,6 +189,22 @@ def _build_refined_op_graph(self):
             op_handle.op_cls.setup_streams(
                 DataStreams(op_handle.input_streams), **op_handle.setup_args)
 
+        # make stream name unique
```
ICGog: This logic should not happen in `graph.py`. IMHO, each data stream should have a uid, which should be deterministically generated (determinism is useful for debugging). Then we'll have to ensure that we use `stream.get_uid()` instead of `stream.name` in the places we index by stream name. Moreover, we won't have to de-unify streams in `*_output_streams` as you do now. What do you think?

Issue #8, opened 2018-11-06
Convert our tests into integration tests


ICGog:
The scripts we have in `tests` are not integration tests. We should convert:

1. `communication_patterns_test.py` into an integration test in which we send n messages, and check that they are received. We should still test the different cases: 1-1, 1-2, 2-1.
2. `control_loop_test.py` into an integration tests that sends n messages in a loop.
3. `ros_integration_tests.py` into integration tests that check if ROSPublisherOp and ROSSubscriberOp work correctly.

PR #13, opened 2018-11-09
[Test] Add basic integration tests

yujialuo:
Addressed https://github.com/erdos-project/erdos/issues/8


tests/communication_1to1_test.py
```diff
@@ -22,7 +24,7 @@
 FLAGS = flags.FLAGS
 flags.DEFINE_string('framework', 'ros',
                     'Execution framework to use: ros | ray.')
-flags.DEFINE_string('case', '1-1', '{num_publisher}-{num_subscriber}, e.g. 1-1, 2-1, 1-2')
+MAXIMUM_COUNT = 5
```
ICGog: Can you please rename to MAX_MSG_COUNT?

tests/communication_1to1_test.py
```diff
@@ -36,6 +38,8 @@ def setup_streams(input_streams):
 
     @frequency(1)
     def publish_msg(self):
+        if self._cnt == MAXIMUM_COUNT:
+            sys.exit()
```
ICGog: Does this work in Ray? I think it crashes the actor. IMHO, we need a different solution here. We need a mechanism to exit an actor. On the publisher side we could do:

```
def execute(self):
  for i in range(0, MAX_MSG_COUNT:
    publish_msg()
    time.sleep(1)
```
We'll have to change `ros_executer` to ensure that we do not spin.

tests/communication_1to1_test.py
```diff
@@ -50,47 +54,40 @@ def execute(self):
...
+            data = msg.data
+        count = int(data.split(" ")[1])
+        assert count < MAXIMUM_COUNT and count == self._cnt, \
+            'received count %d should be equal to %d and smaller than maximum count %d' \
+            % (count, self._cnt, MAXIMUM_COUNT)
+        self._cnt += 1
         print('%s received %s' % (self.name, msg))
 
     def execute(self):
         self.spin()
```
ICGog: Here instead of spinning we could:
```
while self._cnt < MAX_COUNT:
  time.sleep(0.1)
```
We would also have to set a timeout so that the test completes with an error in the case when we don't receive any messages.

tests/communication_1to2_test.py
```diff
@@ -0,0 +1,95 @@
+import os
```
ICGog: Why did you split them into several files? We can have a single communication script file. We can trigger the different test cases using flags.

tests/control_loop_test.py
```diff
@@ -36,13 +40,22 @@ def setup_streams(input_streams):
 
     @frequency(1)
     def publish_msg(self):
+        if self._cnt == MAXIMUM_COUNT:
+            sys.exit()
```
ICGog: As above. I think we need a mechanism to exit from operators.
How can you distinguish between a correct or a failed execution when you invoke `sys.exit`?

PR #14, opened 2018-11-09
[Bug] Fix stream name clashing problem

yujialuo:
Addressed https://github.com/erdos-project/erdos/issues/7

### DIFF
--- a/erdos/data_stream.py
+++ b/erdos/data_stream.py
@@ -63,5 +63,10 @@ class DataStream(object):
         """
         return self.labels.get(key, None)
 
-    def set_uid(self, sender_op_id):
+    @property
+    def uid(self):
+        return self.uid
+
+    @uid.setter
+    def uid(self, sender_op_id):
         self.uid = "{}/{}".format(sender_op_id, self.name)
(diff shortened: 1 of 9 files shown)


erdos/graph.py
```diff
@@ -198,26 +200,26 @@ def _build_output_stream_sinks_graph(self):
         # Assume stream names are unique in the same graph
```
ICGog: Can you please remove this comment?

erdos/graph.py
```diff
@@ -143,6 +143,8 @@ def _build_refined_op_graph(self):
                     output_streams = op_handle.op_cls.setup_streams(
                         DataStreams(current_input_streams),
                         **op_handle.setup_args)
+                    for stream in output_streams:
+                        stream.set_uid(op_id)
```
ICGog: Can you please rename to `set_publisher_id()`? I also think we should use the pythonic way of getting uid ([getters and setters](https://stackoverflow.com/questions/2627002/whats-the-pythonic-way-to-use-getters-and-setters)) so that we can throw an error if stream.uid is None.

erdos/data_stream.py
```diff
@@ -51,11 +53,20 @@ def _copy_stream(self):
...
     def get_label(self, key):
         """
         Retrieves the label for the given key. Returns None if the key
         is not present.
         """
         return self.labels.get(key, None)
+
+    @property
+    def uid(self):
+        return self.uid
```
ICGog: Please raise error/exception if self.uid is None.

PR #17, opened 2018-11-14
Added Window operator.

ICGog:


erdos/operators.py
```diff
@@ -257,3 +260,299 @@ def on_msg(self, msg):
             Message(left_val, Timestamp(msg.timestamp)))
         self.get_output_stream(self._output_stream_name2).send(
             Message(right_val, Timestamp(msg.timestamp)))
+
+
+def get_timestamp_ms():
+    return int(time.time() * 1000)
+
+
+def get_window_start_with_offset(timestamp, window_size_ms, window_offset_ms):
```
pschafhalter: The function doesn't use the `timestamp` argument.

erdos/operators.py
```diff
@@ -257,3 +260,299 @@ def on_msg(self, msg):
...
+    return ((get_timestamp_ms() - window_offset_ms) / window_size_ms *
+            window_size_ms)
```
pschafhalter: This simplifies to `get_timestamp_ms() - window_offset_ms`.

erdos/operators.py
```diff
@@ -257,3 +260,299 @@ def on_msg(self, msg):
...
+    def assign_windows(self, msg):
+        # NOTE: The window assigner assumes that the first coordinate
+        # of the timestamp represents event/source time.
+        start_time = get_window_start_with_offset(msg.timestamp.coordinates[0],
+                                                  self._size, self._offset)
+        return [TimeWindow(start_time, start_time + self._size)]
+
+
+class SlidingWindowAssigner(WindowAssigner):
+    """Asiggns messages to sliding windows.
```
pschafhalter: Typo

erdos/operators.py
ICGog: I don't think it does. `get_timestamp_ms()` returns an int.

pschafhalter: Shouldn't `(...) / window_size_ms * window_size_ms` simplify to (...)?
Unless I'm missing something `window_size_ms / window_size_ms == 1`, right?

Or are you relying on integer division? In that case, use `//` (floor div) instead of `/`, because python 3 will convert integers to floating point if there's a remainder on division.

ICGog: Fixed, I was relying on integer division :).

PR #19, opened 2018-11-15
[Deadline] Basic API

yujialuo:
Added per operator deadline API


erdos/utils.py
```diff
@@ -1,11 +1,44 @@
...
+from threading import Thread
 
 _freq_called = set([])
 
 
+def deadline(*expected_args):
+    """
+    Deadline decorator to be used for restraining computation latency
+    :param duration: computation's duration constrain in ms
+    :param deadline_missing_callback: function to call when the deadline is missed
```
ICGog: The decorator doesn't currently receive a callback, but it should.

erdos/utils.py
```diff
@@ -1,11 +1,44 @@
...
+    :return:
+    """
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            global task_finished, target_ending_time
```
ICGog: I think we shouldn't use global here, but safer solutions like queues. See the [https://stackoverflow.com/questions/19790570/using-a-global-variable-with-a-thread](https://stackoverflow.com/questions/19790570/using-a-global-variable-with-a-thread).

tests/deadline_test.py
```diff
@@ -0,0 +1,105 @@
...
+try:
+    from std_msgs.msg import String
+except ModuleNotFoundError:
+    # ROS not installed
+    String = str
+
+FLAGS = flags.FLAGS
+flags.DEFINE_string('framework', 'ros',
+                    'Execution framework to use: ros | ray.')
+flags.DEFINE_string('case', '1-1',
```
ICGog: Please remove. It is not used.

erdos/utils.py
```diff
@@ -1,11 +1,49 @@
...
+    :param deadline_missing_callback: [str] function name to call when the deadline is missed
+    :return:
+    """
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+
+            def check_deadline(target_ending_time):
+                global task_finished
+                while not task_finished and time.time() < target_ending_time:
```
ICGog: The code here is not thread safe because you're not locking `task_finished`. I checked and it looks like there's a more elegant way to do this using Conditions in Python. The code would look like this:
```
condition = threading.Condition()


def check_deadline(end_time):
    condition.acquire()
    condition.wait(timeout=end_time - time.time())
    condition.release()
    if time.time() >= end_time:
        getattr(args[0], expected_args[1])()

check_thread = threading.Thread(target=check_deadline,
                                args=(target_ending_time,))
check_thread.start()
condition.acquire()
func(*args, **kwargs)
condition.notify()
condition.release()
check_thread.join()
```

PR #22, opened 2018-11-18
Added script to analyze synthetic Pylot logs.

ICGog:


examples/benchmarks/pylot/pylot_log_analyzer.py
```diff
@@ -0,0 +1,107 @@
...
+    def execute(self):
+        output_stream = self.get_output_stream(self._output_name)
+        with open(self._log_file_name, 'r') as log_file:
+            csv_reader = csv.reader(log_file)
+            for row in csv_reader:
+                name = row[0]
+                processing_time = float(row[1])
+                coordinates = [
+                    int(val) for val in row[2].strip('[]').split(',')
+                ]
```
42B: Reading this list comprehension made me think of another option, `ast.literal_eval`

```python
In[17]: s = "[1, 2, 3, 4]"
In[18]: [int(val) for val in s.strip('[]').split(',')]
Out[18]: [1, 2, 3, 4]
In[19]: from ast import literal_eval
In[20]: literal_eval(s)
Out[20]: [1, 2, 3, 4]
```
I still think this list comprehension better adheres to `"Explicit is better than implicit"` though!

PR #26, opened 2018-11-29
[Logging] Operator Level Logging

yujialuo:
Options to log all input streams or all output streams

### DIFF
--- a/erdos/op.py
+++ b/erdos/op.py
@@ -1,5 +1,6 @@
 import sys
 from time import sleep
+import logging
 
 from utils import setup_logging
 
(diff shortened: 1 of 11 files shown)


pschafhalter:
The basic graph API is merged already. The main difference is that you override `Graph.construct` instead of `Graph.setup_streams`.

It should be straightforward to implement -- you can already log input and output streams by adding logging to the graph's input and output operators.

Logging on the internal streams might be a bit more difficult, but should be doable as well.

erdos/op.py
```diff
@@ -86,14 +92,22 @@ def spin(self):
     def log_event(self, processing_time, timestamp, log_message=None):
         pass
 
+    def log_streams(self, stream_uid, msg):
+        self.loggers[stream_uid].info(msg)
```
ICGog: I think we want to `pickle.dump()` here instead of `.info()` so that we can replay the stream we log.

erdos/op.py
```diff
@@ -86,14 +92,22 @@ def spin(self):
...
+
     def _add_input_streams(self, input_streams):
         """Setups and updates all input streams."""
         self.input_streams = self.input_streams + input_streams
+        if self.log_input:
+            for stream in self.input_streams:
+                self.loggers[stream.uid] = setup_logging(stream.uid, log_file="{}.log".format(stream.uid))
```
ICGog: I think we actually want to `pickle.dump()` to a file rather than use a logger. If we use loggers, we won't be able to replay the streams.

ICGog:
`RecordOp` and `ReplayOp` are good examples on how to log & replay.

erdos/op.py
```diff
@@ -86,14 +92,22 @@ def spin(self):
     def log_event(self, processing_time, timestamp, log_message=None):
         pass
 
+    def record_streams(self, stream_uid, msg):
```
ICGog: IMHO, recording messages of a stream should not happen in the operator. Recording a stream is an activity that is strictly associated with a stream. If we follow the code path, we now invoke `record_streams` from the stream, and we even pass the message for the recording to happen. I would move this logic into the data streams. As a result, we would also handle for free cases such as recording only once data streams that multiple receivers.

erdos/op.py
```diff
@@ -86,14 +92,22 @@ def spin(self):
...
+        pickle.dump(msg, self.recorders[stream_uid])
+
     def _add_input_streams(self, input_streams):
         """Setups and updates all input streams."""
         self.input_streams = self.input_streams + input_streams
+        if self.record_input:
+            for stream in self.input_streams:
```
ICGog: With the change from above, this logic would only set the stream as a recording one.

erdos/utils.py
```diff
@@ -76,6 +77,8 @@ def setup_logging(name, log_file=None):
     if log_file is None:
         handler = logging.StreamHandler()
     else:
+        if '/' in log_file:  # Slash will be mistaken as path path in logger
```
ICGog: I think this is dangerous behavior because we're changing log_file implicitly. IMHO, we should provide an auxiliary method that developers can use if they want (e.g., `escape_log_file`), and raise an exception here if the log_file is not suitable.

erdos/utils.py
```diff
@@ -114,3 +117,13 @@ def log_graph_to_dot_file(filename, nodes, edges):
     # dot footer
     dot_file.write("}")
     dot_file.close()
+
+
+def setup_recorder(name, record_type):
+    filename = '{}.pkl'.format(name)
+    if '/' in name:  # Slash will be mistaken as path path
+        filename = '_'.join(filename.split('/'))
```
ICGog: As above.

erdos/utils.py
```diff
@@ -114,3 +117,13 @@ def log_graph_to_dot_file(filename, nodes, edges):
...
+    f = open(filename, "wb")
+    record_info = "record name: {}; record type: {}".format(name, str(record_type))
+    pickle.dump(record_info, f)
```
ICGog: One issue here is that we never close the recorded file, which means we may not record the last few messages. Instead, we should ensure that the file is closed when the stream is destructed. IMHO, given that the recorded file is strictly a property of the stream, this whole setup_recorder method should be in the data_stream class, and the file should be a field.

PR #31, opened 2019-03-26
[IMPL] Implemented an EarlyReleaseOperator.

sukritkalra:
- Added functions to ensure that `len(...)`, iteration and indexing works on `DataStreams`.
- Changed the ROS and Ray implementation to only flow watermarks if the operator does not define a single callback on a stream. We assume that if a callback is defined, the developer flows the watermark correctly.
- Implemented an `EarlyReleaseOperator` with the following semantics:
  - Forward messages from the first stream that releases a message corresponding to a particular timestamp.
  - Release messages and watermarks for the same timestamp only from that stream.


erdos/ray/ray_operator.py
```diff
@@ -52,15 +52,46 @@ def on_completion_msg(self, msg):
...
+
+        # Now check if all other streams have a higher or equal watermark.
+        # If yes, flow this watermark. If not, return from this function
+        # Also, maintain the lowest watermark observed.
+        low_watermark = msg.timestamp
+        for stream, watermark in self._op._stream_to_high_watermark.items():
+            if stream != msg.stream_name:
+                if not watermark or watermark < msg.timestamp:
+                    return
+                if low_watermark > watermark:
```
ICGog: Doesn't this work correctly when you have two input streams (i1 and i2), i2 has received watermark t + 1, and i1 just received watermark t? If I understand the code correctly, it will update low_watermark to t +1, and send watermark t + 1, instead of t.

ICGog: I take it back :)

ICGog: I think there's a case here when you'll end up duplicating watermarks. Input stream i1 received watermark t, i2 received watermark t (and you issued downstream watermark t). Now if i2 receives t +1, you'll send again downstream watermark t.

IMHO, the code would be easier to read if you kept a global low watermark, do a loop over the watermarks to get the min watermark on the inputs. If this min is greater than the global, update and send, otherwise return.

erdos/ros/ros_input_data_stream.py
```diff
@@ -36,6 +40,35 @@ def _on_msg(self, msg):
...
+                        "on the same stream: {}".format(msg, high_watermark))
+                else:
+                    self.op._stream_to_high_watermark[msg.stream_name] = \
+                            msg.timestamp
+
+            # Now check if all other streams have a higher or equal watermark.
+            # If yes, flow this watermark. If not, return from this function
+            # Also, maintain the lowest watermark observed.
+            low_watermark = msg.timestamp
+            for stream, watermark in self.op._stream_to_high_watermark.items():
```
ICGog: Same as above.

tests/watermarks_block.py
```diff
@@ -0,0 +1,122 @@
...
+            "[{}] Received the message with data {} and timestamp {}.".format(
+                self.name, msg.data, msg.timestamp.coordinates[0]))
+        self.timestamp_to_data_map[msg.timestamp.coordinates[0]].append(
+            msg.data)
+
+    def on_watermark(self, msg):
+        print("[{}] Received the watermark for timestamp {}.".format(
+            self.name, msg.timestamp.coordinates[0]))
+        data_values = self.timestamp_to_data_map[msg.timestamp.coordinates[0]]
+        self.timestamp_to_data_map[msg.timestamp.coordinates[0]] = []
```
ICGog: Can you add an assert here to check we don't receive duplicate watermarks and that they're received in increasing order?

erdos/ray/ray_operator.py
sukritkalra: I don't think the situation you pointed out is possible. The `if not watermark or watermark < msg.timestamp` protects against that.
So, when i1 receives a watermark t, and i2 has no initial watermark, the `if not watermark` clause will be invoked and the method will return.
When i2 receives a watermark t, i1 will have the watermark t and the `not watermark or watermark < msg.timestamp` will evaluate to `False`, and thus the timestamp t will flow.
When i2 receives a watermark t+1, `watermark < msg.timestamp` will evaluate `True` for i1 and we will not flow a watermark downstream.

PR #32, opened 2019-04-02
[Ray] Schedule operators on specific machines by setting resources

pschafhalter:
To schedule an operator on a machine:
1. Initialize the machine with a custom resource, e.g. `ray.start --resources="{'machine_a': 100}"
2. Require that machine's custom resource when initializing the operator, e.g. `erdos.add(MyOp, resources={"machine_a": 1})

I also removed dead code when we had a `machine` argument for scheduling since custom resources are more general and less maintenance for now.

### DIFF
--- a/erdos/ray/ray_executor.py
+++ b/erdos/ray/ray_executor.py
@@ -16,10 +16,9 @@ class RayExecutor(Executor):
     def setup(self):
         resources = dict(
             self.op_handle.resources) if self.op_handle.resources else {}
-        if self.op_handle.machine:
-            resources[self.op_handle.machine] = 1
         num_cpus = self.op_handle.resources.pop("CPU", None)
         num_gpus = self.op_handle.resources.pop("GPU", None)
+        print(self.op_handle.resources)
 
         # TODO (Yika): hacky solution for using decorator on callbacks
         # When we wrap op in ray operator, __name__ of callbacks that have
--- a/erdos/ray/ray_executor.py
+++ b/erdos/ray/ray_executor.py
@@ -18,7 +18,6 @@ class RayExecutor(Executor):
             self.op_handle.resources) if self.op_handle.resources else {}
         num_cpus = self.op_handle.resources.pop("CPU", None)
         num_gpus = self.op_handle.resources.pop("GPU", None)
-        print(self.op_handle.resources)
 
         # TODO (Yika): hacky solution for using decorator on callbacks
         # When we wrap op in ray operator, __name__ of callbacks that have
--- a/tests/ray_cluster_test.py
+++ b/tests/ray_cluster_test.py
@@ -29,7 +29,7 @@ class MachineAOp(Op):
     def print_node(self):
         plasma_store_socket_name = (
             ray.worker.global_worker.plasma_client.store_socket_name)
-        node_name = node_directory[plasma_store_socket_name]
+        node_name = node_directory.get(plasma_store_socket_name, "unknown")
         print("MachineAOp is located on node {}".format(node_name))
 
     def execute(self):
@@ -49,7 +49,7 @@ class MachineBOp(Op):
     def print_node(self):
         plasma_store_socket_name = (
             ray.worker.global_worker.plasma_client.store_socket_name)
-        node_name = node_directory[plasma_store_socket_name]
+        node_name = node_directory.get(plasma_store_socket_name, "unknown")
         print("MachineBOp is located on node {}".format(node_name))
 
     def execute(self):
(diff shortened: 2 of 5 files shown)


tests/erdos.gv
```diff
@@ -0,0 +1 @@
+digraph G {	graph [rankdir="LR"]	{ node [shape=box]	sum_squares_square [ label = "sum_squares_square" ];default_sum_squares_output_op [ label = "default_sum_squares_output_op" ];sum_squares_sum [ label = "sum_squares_sum" ];default_int2 [ label = "default_int2" ];default_int1 [ label = "default_int1" ];default_sum_squares_input_op [ label = "default_sum_squares_input_op" ];default_default_square [ label = "default_default_square" ];	}	{ edge [color="#cccccc"]	default_int2 -> default_sum_squares_input_op [ label = "" ];default_sum_squares_input_op -> sum_squares_square [ label = "" ];sum_squares_square -> sum_squares_sum [ label = "" ];sum_squares_sum -> default_sum_squares_output_op [ label = "" ];default_sum_squares_output_op -> default_default_square [ label = "" ];default_int1 -> default_sum_squares_input_op [ label = "" ];	}}
```
sukritkalra: Can you please remove this file from the commit? Same for all the other binary swp files.

tests/ray_cluster_test.py
```diff
@@ -0,0 +1,88 @@
+from __future__ import print_function
+
+from absl import app
+from absl import flags
+
+from ray.tests.cluster_utils import Cluster
```
sukritkalra: ERDOS abstractions should not leak anything about the underlying execution system, except when `graph.execute()` is called. 

We should also not depend on test utils, since it is not a stable and documented API.

erdos/ray/ray_executor.py
```diff
@@ -16,10 +16,9 @@ def __init__(self, op_handle):
     def setup(self):
         resources = dict(
             self.op_handle.resources) if self.op_handle.resources else {}
-        if self.op_handle.machine:
-            resources[self.op_handle.machine] = 1
         num_cpus = self.op_handle.resources.pop("CPU", None)
         num_gpus = self.op_handle.resources.pop("GPU", None)
+        print(self.op_handle.resources)
```
sukritkalra: Remove the `print`.

tests/ray_cluster_test.py
```diff
@@ -0,0 +1,88 @@
...
+                            resources={"machine_a": 1})
+    machine_b_op = graph.add(MachineBOp, name='machine_b_op',
+                            resources={"machine_b": 1})
+
+    # Execute graph
+    graph.execute("ray")
+
+
+if __name__ == "__main__":
+    cluster = Cluster(initialize_head=True)
```
sukritkalra: We would ideally want to have a machine and cluster abstraction that is independent of the underlying execution platform. The system needs to be able to distribute the graph both using `ROS` and `Ray` cleanly.

I would envision writing a `Machine` and a `Cluster` class that gets passed to a `GraphNode` or a `Graph` respectively, and the initialization of the execution engine happens during the setup phase. (creation of custom resources and allocation of those resources per node for Ray, and setup of a ROS master on the master node and an export of the `ROS_MASTER_URI` on every other node.)

tests/ray_cluster_test.py
```diff
@@ -0,0 +1,88 @@
...
+    a = cluster.add_node(resources={"machine_a": 100}, num_cpus=4)
+    b = cluster.add_node(resources={"machine_b": 100}, num_cpus=4)
+
+    node_directory[a.plasma_store_socket_name] = "machine_a"
```
sukritkalra: Are `machine_a` and `machine_b` assumed to be host names in `/etc/hosts`? How does one start the Ray nodes there without SSHing into the machine and doing a `ray init`?

tests/ray_cluster_test.py
```diff
@@ -0,0 +1,88 @@
+from __future__ import print_function
+
+from absl import app
+from absl import flags
+
+from ray.tests.cluster_utils import Cluster
+
+import ray.worker
```
sukritkalra: We would ideally like to remove any execution backend specific imports.

tests/ray_cluster_test.py
pschafhalter: I'm not really sure how else to quickly simulate a Ray cluster without manually SSHing into several different nodes. The Ray cluster abstraction is reliable and widely used across Ray tests. Do you have any suggestions on what could be better?

Anyway, doesn't the `--ray_redis_address` flag already leak information about the underlying execution system?

tests/ray_cluster_test.py
pschafhalter: This simulates a cluster on 1 node.

PR #33, opened 2019-04-04
Cluster support

pschafhalter:
Adds cluster abstractions to ERDOS for Ray and ROS.


erdos/cluster.py
```diff
@@ -0,0 +1,82 @@
...
+
+        if setup_env_commands is not None:
+            for command in setup_env_commands:
+                node.run_command(command)
+
+        self.setup_node(node)
+
+        return node
+
+    def setup_node(self, node):
```
ICGog: This should be called `setup` and be invoked to setup the cluster. Internally, the method may invoke `setup` on each node.

erdos/cluster.py
```diff
@@ -0,0 +1,82 @@
...
+    def connect(self):
+        """Creates an SSH connection to the node.
+
+        This is called by default in __init__.
+        """
+        if not self.connection.login(
+                self.server, self.username, ssh_key=self.ssh_key):
+            raise Exception("Unable to log into {} with username {}".format(
+                self.server, self.username))
+
```
ICGog: Should have a `setup` method.

erdos/ray/ray_cluster.py
```diff
@@ -0,0 +1,55 @@
...
+        response = node.run_command(ray_start_command).decode("utf-8")
+
+        matches = re.findall('redis_address="[0-9.:]+"', response)
+        if len(matches) != 1:
+            raise Exception("Unable to parse redis address")
+
+        self.redis_address = matches[0].split('"')[1]
+
+    def setup_worker(self, node):
+        resources = {node.name: 512}
```
ICGog: Do all the nodes have the same 512 of what? I'm not sure I understand this line.

erdos/ray/ray_cluster.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, redis_address="", can_install_ray=False):
+        super(RayCluster, self).__init__()
+        self.redis_address = redis_address
+        self.can_install_ray = can_install_ray
+
+    def __del__(self):
+        self.broadcast_command("ray stop")
+        super(RayCluster, self).__del__()
+
+    def install_ray(self, node):
```
ICGog: This is great, but we don't have to worry about installing the dependency on the workers in the v0. We can assume that we have all the ROS/Ray dependencies on each node.

erdos/cluster.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+
+    def __init__(self):
+        self.nodes = {}
+
+    def __del__(self):
+        for node in self.nodes.values():
+            node.disconnect()
+
+    def add_node(self,
```
ICGog: IMHO, we should plug out the node construction from `add_node`. Instead the method should receive a node object. In this way, we avoid coupling nodes with clusters, we can enrich nodes with more info (e.g., resources), and use node objects to do scheduling for ROS.

erdos/ray/ray_cluster.py
pschafhalter: Ray can't assign more than 512 of a resource to a node.

erdos/cluster.py
pschafhalter: Sounds good! I was struggling with this, and decided to implement a top-down approach, where nodes are the same and specialized clusters invoke different commands on nodes.

If we're looking for more fine-grained control on individual nodes, enriching nodes makes sense.

pschafhalter:
@ICGog this should be more in line with the abstractions we discussed on Friday. I'm struggling a bit with the tension between the executor and `Cluster.execute`, especially the `RayExecutor`.

I'm thinking of moving the logic for `RayExecutor.execute()` to `RayCluster.execute()`, so that `RayExecutor.execute()` just calls `cluster.execute()`. Do you think this is a good approach or do you have another suggestion?

pschafhalter:
I agree that those changes make sense. Thanks for the feedback!

erdos/local/local_dispatcher.py
```diff
@@ -0,0 +1,33 @@
...
+
+        start = time.time()
+        end = start + timeout
+        while time.time() < end:
+            if not self.current_process.isalive():
+                process = self.current_process
+                self.current_process = None
+                self._cleanup_processes()
+                return process.read()
+
```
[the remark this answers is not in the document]
pschafhalter: Do you mean if the timeout is violated? Because this checks if the process completes, and if it does, prints all the output which is similar to the command line behavior.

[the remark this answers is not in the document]
ICGog: Yes, if the timeout is violated (i.e., the process doesn't complete in time).

I just noticed that this busy loop doesn't have a time.sleep. Should we do a time.sleep(smallish_value) so that we give the CPU core some time to do other work? Otherwise, this while loop will use an entire core.

PR #34, opened 2019-04-08
Flux

yujialuo:
- Fault Tolerance under Normal Operation [DONE]
- Take over upon failure [DONE]
- Catch up and recover [IN PROGRESS]

To run the example: 
`python active_replication/flux_test.py --framework=ros`


erdos/graph.py
```diff
@@ -263,7 +263,8 @@ def _build_refined_op_graph(self):
         for op_id, op_handle in self.op_handles.items():
             op_handle.op_cls.setup_streams(
                 DataStreams(op_handle.input_streams), **op_handle.setup_args)
-
+        for op_id, op_handle in self.op_handles.items():
+            print(op_handle.name, [out.name for out in op_handle.output_streams])
```
ICGog: Can you please improve logging here? Please give the context of what you're logging (e.g., Operator foo has output streams: bar, bar2)

examples/active_replication/controller_operator.py
```diff
@@ -0,0 +1,47 @@
...
+from erdos.timestamp import Timestamp
+from erdos.utils import setup_logging
+
+import flux_utils
+
+
+class ControllerOperator(Op):
+    def __init__(self,
+                 name,
+                 fail_timeout=1,
```
ICGog: Can you please add a time unit? (e.g., fail_timeout_ms = 1)?

examples/active_replication/controller_operator.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def execute(self):
+        # TODO(ionel): We should send the failure message on a failed stream,
+        # and get a reply back so that we know the node is not processing or
+        # issuing any new messages. Otherwise, we run into the risk of having
+        # inconsistent state because we can't guarantees in-order delivery
+        # across operators.
+        
+        # Send failure message.
+        time.sleep(3)
```
ICGog: Can you please parametrize the time after which to insert a failure using an operator parameter?

examples/active_replication/controller_operator.py
```diff
@@ -0,0 +1,47 @@
...
+        # Fail primary and notify ingress, egress and consumers
+        fail_replica_num = 0
+        fail_msg = Message((flux_utils.FluxControllerCommand.FAIL, fail_replica_num), Timestamp(coordinates=[0]))
+        self.get_output_stream('controller_stream').send(fail_msg)
+        print("Control send failure message to primary")
+
+        # Recover failed primary
+        time.sleep(self._timeout)
```
ICGog: Oh, this is a failure_duration. Can you please rename self._timeout to self._failure_duration and likewise fail_timeout.

examples/active_replication/failure_operator.py
```diff
@@ -0,0 +1,46 @@
...
+    @staticmethod
+    def setup_streams(input_streams, output_stream_name):
+        input_streams.filter(is_flux_consumer_output).add_callback(
+            FailureOperator.on_msg)
+        input_streams.filter(is_control_stream).add_callback(
+            FailureOperator.on_control_msg)
+        return [DataStream(name=output_stream_name)]
+
+    def on_msg(self, msg):
+        if not self._failed:
```
ICGog: Can you please add a comment here? Please explain that we're not processing messages if we've declared this node as failed.

examples/active_replication/failure_operator.py
```diff
@@ -0,0 +1,46 @@
...
+            self.get_output_stream(self._output_stream_name).send(msg)
+
+    def on_control_msg(self, msg):
+        (control_num, replica_num) = msg.data
+        if replica_num != self._replica_num:
+            pass
+        elif control_num == flux_utils.FluxControllerCommand.FAIL:
+            self._logger.info("Failed by controller.")
```
ICGog: Please include the replica name in the log as well.

examples/active_replication/failure_operator.py
```diff
@@ -0,0 +1,46 @@
...
+            self._failed = True
+        elif self._failed and control_num == flux_utils.FluxControllerCommand.RECOVER:
+            self._failed = False
```
ICGog: Please log here.

examples/active_replication/sink.py
```diff
@@ -0,0 +1,20 @@
...
+        super(Sink, self).__init__(name)
+        self._logger = setup_logging(self.name, log_file_name)
+
+    @staticmethod
+    def setup_streams(input_streams):
+        input_streams.filter(is_not_ack_stream).add_callback(Sink.on_msg)
+        return []
+
+    def on_msg(self, msg):
+        print('%s received %s' % (self.name, msg))
```
ICGog: Can you please use asserts here to check that you're not missing messages and also not receiving duplicates?

examples/active_replication/source.py
```diff
@@ -0,0 +1,26 @@
...
+        self._logger = setup_logging(self.name, log_file_name)
+        self._cnt = 0
+        self._seq_num = 0
+
+    @staticmethod
+    def setup_streams(input_streams):
+        return [DataStream(name='input_stream')]
+
+    def execute(self):
+        while self._seq_num < 50:
```
ICGog: Please use self._num_messages instead of 50, and pass num_messages in the constructor with a default value of 50.

examples/active_replication/source.py
```diff
@@ -0,0 +1,26 @@
...
+            output_msg = Message("data"+str(self._seq_num),
+                                 Timestamp(coordinates=[self._seq_num]))
+            self.get_output_stream('input_stream').send(output_msg)
+            self._seq_num += 1
+            time.sleep(0.1)
```
ICGog: Please make the frequency a parameter to the constructor, so that we can easily test different frequencies.

erdos/graph.py
yujialuo: Oops that was for debugging forgot to delete it

examples/active_replication/flux_buffer.py
```diff
@@ -0,0 +1,54 @@
...
+            if self.queue[i][0] == sn and self.queue[i][2] == dest:
+                self.queue.pop(i)
+                return True
+        return False
+
+    def ack_all(self, dest):
+        i = 0
+        while i < len(self.queue):
+            if self.queue[i][2] == dest:
+                self.queue.pop(i)
```
ICGog: I see now why you're queuing a message for each replica. The approach I proposed above is not sufficient to implement this method. However, you can still avoid duplicating every data message in the queue by stroing (seq_num, data, list_of_non_acknowledge_replicas).

examples/active_replication/flux_buffer.py
```diff
@@ -0,0 +1,54 @@
...
+            else:
+                i += 1
+
+    def reset(self):
```
ICGog: I see this method is not used anywhere. Is it here for the recovery case? Could you please add a comment to explain what it does because it is not clear to me.

examples/active_replication/flux_buffer.py
```diff
@@ -0,0 +1,54 @@
...
+        self.queue.pop(i)
+
+    def size(self):
+        return len(self.queue)
+
+    def match_oldest(self, sn):
+        return self.queue[0][0] == sn
+
+    def pop_oldest(self):
+        self.queue.pop(0)
```
ICGog: Please exception if queue is empty.

examples/active_replication/flux_buffer.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def send_and_clear(self, pub):
```
ICGog: This is a dangerous method given the way the buffer stores data (the same message may be stored twice). IMHO, we should only expose `pop(dest)` which returns the first message. We can use this method externally to send the data.

examples/active_replication/flux_consumer_operator.py
```diff
@@ -0,0 +1,63 @@
...
+                 output_stream_name,
+                 ack_stream_name,
+                 log_file_name=None):
+        super(FluxConsumerOperator, self).__init__(name)
+        self._logger = setup_logging(self.name, log_file_name)
+        self._replica_num = replica_num
+        self._output_stream_name = output_stream_name
+        self._ack_stream_name = ack_stream_name
+        self._failed = False
+        self.lock = threading.Lock()
```
ICGog: Please make lock "private".

examples/active_replication/flux_ingress_operator.py
```diff
@@ -0,0 +1,77 @@
...
+                 name,
+                 output_stream_names,
+                 num_replics=2,
+                 log_file_name=None):
+        super(FluxIngressOperator, self).__init__(name)
+        self._logger = setup_logging(self.name, log_file_name)
+        self._num_replicas = num_replics
+        self._output_streams = output_stream_names
+        self._input_msg_seq_num = 0
+        self.buffer = Buffer(num_replics)   # buffer to store unacknowledged tuples
```
ICGog: Please make buffer "private".

examples/active_replication/flux_ingress_operator.py
```diff
@@ -0,0 +1,77 @@
...
+        self._status = {}
```
ICGog: Why is this a set if we assume replicas have sequence numbers as ids? It could be a list. Is this for the future when we may have an unique id for each replica?

examples/active_replication/flux_ingress_operator.py
```diff
@@ -0,0 +1,77 @@
...
+            .add_callback(FluxIngressOperator.on_msg)     # Input
+        input_streams.filter(is_ack_stream).add_callback(
+            FluxIngressOperator.on_ack_msg)     # Ack
+        input_streams.filter(is_control_stream).add_callback(
+            FluxIngressOperator.on_control_msg)     # Control
+
+        return [DataStream(name=output_stream_names)]
+
+    def on_msg(self, msg):
+        self.lock.acquire()
```
ICGog: Could you please explain why are you locking here? Are you assuming that on_msg can be called simultaneously by multiple threads?

examples/active_replication/flux_ingress_operator.py
```diff
@@ -0,0 +1,77 @@
...
+from flux_utils import is_ack_stream, is_control_stream, is_not_ack_stream, is_not_control_stream, is_not_back_pressure
+from flux_buffer import Buffer
+import flux_utils
+import threading
+
+
+class FluxIngressOperator(Op):
+    def __init__(self,
+                 name,
+                 output_stream_names,
```
ICGog: Please rename to output_stream_name.

examples/active_replication/flux_ingress_operator.py
```diff
@@ -0,0 +1,77 @@
...
+                 num_replics=2,
+                 log_file_name=None):
+        super(FluxIngressOperator, self).__init__(name)
+        self._logger = setup_logging(self.name, log_file_name)
+        self._num_replicas = num_replics
+        self._output_streams = output_stream_names
```
ICGog: Likewise, output_stream_name

examples/active_replication/flux_consumer_operator.py
```diff
@@ -0,0 +1,63 @@
...
+            self.get_output_stream(self._ack_stream_name).send(
+                Message((self._replica_num, msg_seq_num), msg.timestamp))
+            # 2) Forward the message
+            self.get_output_stream(self._output_stream_name).send(msg)
+
+    def on_control_msg(self, msg):
+        self.lock.acquire()
+        (control_num, replica_num) = msg.data
+        if replica_num != self._replica_num:
+            pass
```
ICGog: It's safer to return here.

examples/active_replication/flux_producer_operator.py
```diff
@@ -0,0 +1,93 @@
...
+class FluxProducerOperator(Op):
+    def __init__(self,
+                 name,
+                 replica_num,
+                 output_stream_name,
+                 log_file_name=None):
+        super(FluxProducerOperator, self).__init__(name)
+        self._logger = setup_logging(self.name, log_file_name)
+        self._replica_num = replica_num
+        self._ex_secondary = False
```
ICGog: Please add a comment to explain ex_secondary's role.

examples/active_replication/flux_producer_operator.py
```diff
@@ -0,0 +1,93 @@
...
+
+        return [DataStream(name=output_stream_name)]
+
+    def on_msg(self, msg):
+        self.lock.acquire()
+        # XXX(ionel): Need to ensure that we remove output_seq_num
+        # during recovery.
+        if self._replica_num == 0:  # Primary
+            msg.data = (self._output_seq_num, msg.data)
+            if self._ex_secondary:  # used to a secondary
```
ICGog: Will this code ever get invoked because it is in the self._replica_num == 0 body. I think it's safer to have another attribute self._master which you can set. In this way, you won't make the assumption that the master always has replica_num = 0.

ICGog: Also, might be worthwhile to push all this recovery code into a separate method.

examples/active_replication/flux_producer_operator.py
```diff
@@ -0,0 +1,93 @@
...
+                if len(self._ack_buffer) > 0 and max(self._ack_buffer) > self._output_seq_num:   # behind progress
+                    lst = [str(a) for a in self._ack_buffer]
+                    lst = ",".join(lst)
+                    assert self._output_seq_num in self._ack_buffer, str(self._output_seq_num) + " not in " + lst
+                    self._ack_buffer.remove(self._output_seq_num)
```
ICGog: Looking at this code, I think it's much easier if you transform ack_buffer into a queue. ACK_buffer will store the ACKS in increasing order, and we will also always remove the first ACK because we'll also  receive messages in order from the upstream operator.

examples/active_replication/flux_buffer.py
```diff
@@ -0,0 +1,54 @@
+
+class Buffer:
+    def __init__(self, n_dest):
+        self.n_dest = n_dest
+        self.queue = list()  # TODO(yika): add upper bound on buffer size
+
+    def put(self, data, sn, dest):
```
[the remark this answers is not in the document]
yujialuo: Based on the implementation in the paper, same data delivered to different dest are considered different data stored in ingress buffer, so that ack from different dest will drop different data in the buffer

examples/active_replication/flux_buffer.py
```diff
@@ -0,0 +1,54 @@
...
+        # Put input stream
+        if len(self.queue) > 0:
+            assert sn >= self.queue[-1][0]
```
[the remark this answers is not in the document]
yujialuo: Same comment as above: there can be two identical sequence number but different destination. e.g. (4, data4, 1), (4, data4, 2), (5, data5, 1), (5, data5, 2)...

examples/active_replication/flux_buffer.py
yujialuo: this was my previous implementation, but then I realize the paper does it the current way so I figure I should follow it. Also it's more wordy to implement that because tuple is immutable so it's better to do only checking and dropping the whole tuple instead of modifying it

ICGog: Ok, you don't actually duplicate data because you store references in the queue. Then this is fine. It's just that the design is less brittle if you don't assume that replicas have ids starting from zero. 

For example, imagine a setup in which the data-flow runs for a long time, while nodes fail and come back. After every single failure, you would have to ensure that the replica that replaces the failed one gets the id of the failed one. If you don't make the assumption that ids are sequential, and starting from zero, you don't have to do anything (wrt to the data structure) when you bring up a new replica.

examples/active_replication/flux_buffer.py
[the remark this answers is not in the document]
ICGog: Yep, I realized that when I read the ack_all method. However, you can store this as (4, data4, [replica_id1, replica_id2]). In this way, the code is not going to make any assumptions about replica ids. Please see third comment below.

examples/active_replication/flux_ingress_operator.py
yujialuo: oh that means a dictionary. set is declared as a = set()

examples/active_replication/flux_ingress_operator.py
yujialuo: hmm my understanding is that on_msg and on_ack_msg will both change state of the op, but we want any change of state to be atomic, so this is preventing on_msg and on_ack_msg try to change state at the same time.

examples/active_replication/flux_buffer.py
ICGog: Re immutable: Good point, however, the tuple stores a reference to the list. Hence, you don't modify the tuple, only the list it references to.

Ignore my comment on the same id.

Ok let's leave it as it is, but please add a comment to explain the alternative option.

examples/active_replication/flux_consumer_operator.py
yujialuo: Ah I need to release the lock afterwards so...

examples/active_replication/flux_ingress_operator.py
ICGog: Ignore my comment. Let's assume replicas can have non-sequential ids => dict works great.

examples/active_replication/flux_ingress_operator.py
ICGog: I think this case cannot happen with our current ROS & Ray implementations:
1. Ray invokes the callback sequentially
2. Similarly, ROS uses a single thread to invoke them because we use rospy.spin() (see [http://wiki.ros.org/roscpp/Overview/Callbacks%20and%20Spinning](url))

Let's leave the locks in for now, but please add a comment explaining that we're looking in case in the future we'll change the ROSExecutor to use a multi threaded spinner.

examples/active_replication/flux_producer_operator.py
yujialuo: I feel like self._replica_num is completely equivalent to self._master.

ICGog: Imagine you have a primary (replica_num == 0), and a secondary (replica_num == 1). If the primary fails and the secondary becomes primary then the primary will be the node with replica_num == 1. I don't think they're equivalent.

Yes, you do set the replica_num to 0 when you detect that the primary has failed, but IMHO it's easy to introduce bugs if you do that. What if any buffer uses replica num (id), or if any upstream/downstream operator works with replica id. It's just dangerous to change replica id in flight.

examples/active_replication/flux_buffer.py
```diff
@@ -0,0 +1,54 @@
...
+        self.queue.append((sn, data, dest))
+
+    def ack(self, sn, dest):
+        for i in range(len(self.queue)):
+            if self.queue[i][0] == sn and self.queue[i][2] == dest:
```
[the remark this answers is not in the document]
yujialuo: Ok I will add that to comment for future implementation

examples/active_replication/flux_producer_operator.py
yujialuo: Ok will change it. But it will just be renaming self._replica_num to self._primary because I only use self._replica_num for this purpose.

ICGog: Yep, that should do it, if you rename it to self._primary and make it a boolean.

PR #35, opened 2019-04-11
[Fault Tolerance] Checkpointing

yujialuo:
`python checkpoint/checkpoint_test.py --framework=ros`


erdos/ros/ros_input_data_stream.py
```diff
@@ -61,13 +64,20 @@ def _on_msg(self, msg):
...
+                        if not watermark or watermark < msg.timestamp:
+                            return
+                        if low_watermark > watermark:
+                            low_watermark = watermark
             msg = WatermarkMessage(low_watermark)
-            
+
+            # Checkpoint
+            if self.op._checkpoint_enable:
+                if msg.timestamp.coordinates[0] % self.op._checkpoint_freq == 0:
```
ICGog: Can we please call here a user provided lambda instead of hardcoding logic that check on the first timestamp coordinate? The lambda receives msg.timestamp, and thus the user write custom checkpointing policies. For example, in Pylot the timestamps are (simulation_time, sequence_number), and it could be beneficial to be able to use the sequence number to decide when to checkpoint.

erdos/op.py
```diff
@@ -87,6 +92,16 @@ def spin(self):
     def log_event(self, processing_time, timestamp, log_message=None):
         pass
 
+    def checkpoint(self):
+        """ User override """
+        pass
+
```
ICGog: Can you please add a restore method as well?

erdos/ros/ros_input_data_stream.py
```diff
@@ -61,13 +64,20 @@ def _on_msg(self, msg):
             # Also, maintain the lowest watermark observed.
             low_watermark = msg.timestamp
```
ICGog: Does this work when we're recovering? I'm thinking of the case when we rollback all operators to time t, but ROS still stores in its queues watermarks with a higher time (e.g., t + 2). These watermarks will be delivered after operator recovery, and thus we may end up marking time t + 2 as complete before we should have (e.g., after we received the messages tagged with t + 1 and  t + 2 from the recovered operators).

examples/checkpoint/source.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+    @staticmethod
+    def setup_streams(input_streams):
+        input_streams\
+            .filter(checkpoint_util.is_control_stream)\
+            .add_callback(Source.on_rollback_msg)
+        return [DataStream(name='input_stream')]
+
+    def on_rollback_msg(self, msg):
```
ICGog: There's a lot of repetition in these on_rollback_msg methods. Let's discuss in person to see if we can figure a good way to reduce the repetition.

yujialuo:
> Great work! Can you please have a look at my comments? Can you please also add a test that checks for exactly once message delivery? You can implement it using watermark callback.

I already checked at the Sink about exactly-once delivery

PR #36, opened 2019-04-15
[Fault Tolerance] Upstream 

yujialuo:


examples/upstream/upstream_operator.py
```diff
@@ -0,0 +1,67 @@
+from erdos.data_stream import DataStream
+from erdos.op import Op
+from erdos.utils import setup_logging
+from collections import deque
+import upstream_util
+
+
+class UpstreamOperator(Op):
```
ICGog: Can you please generalize the operator (e.g., do not assume a single data stream, do not hard encode output stream names). The code should be written such that we can move it to erdos/operators.py, and we can re-use it whenever we need it.

ICGog:
Closing the issue because we're migrating to a new execution engine.

PR #37, opened 2019-04-19
[Fault Tolerance] Move Checkpointing Logic Into Base Operator Class

yujialuo:


erdos/op.py
```diff
@@ -32,11 +33,16 @@ def __init__(self, name, checkpoint_enable=False, checkpoint_freq=None):
...
+
+        # Checkpoint variables
         self._checkpoint_enable = checkpoint_enable
         self._checkpoint_freq = checkpoint_freq
         if self._checkpoint_enable:
             assert self._checkpoint_freq is not None
-        self._stream_ignore_watermarks = set()  # input streams that do not send watermarks
+        # XXX(ionel): _checkpoints could be a queue, but we kept it a dict
+        # to be able to easily do asserts.
+        self._checkpoints = dict()
```
ICGog: It looks to me that old checkpoints are never garbage collected. IMHO, we need a method in op.py that garbage collects checkpoints up to a timestamp. I'll add in a separate commit.

erdos/op.py
```diff
@@ -121,7 +124,32 @@ def restore(self, timestamp):
...
+        for id in pop_ids:
+            self._checkpoints.pop(id)
+
+        # Reset watermark
+        self._reset_progress(Timestamp(coordinates=[checkpoint_id]))
+
+        return self._checkpoints[checkpoint_id]
+
+    def _checkpoint(self, checkpoint_id):
+        checkpoint_id = checkpoint_id.coordinates[0]
```
ICGog: Why not use the entire timestamp as checkpoint id?

PR #59, opened 2020-01-27
Fix race condition during setup

pschafhalter:
Fixes the race condition where operators may begin executing before all nodes are connected and all operators are ready to run.

The following protocol on the control plane fixes the race condition:
1. The node executor spawns a green thread for each operator.
2.  Each operator initializes by calling `Operator::new()` and sends the `OperatorInitialized` message to the node executor.
3. The node executor waits until it has received `OperatorInitialized` messages from all operators. It then broadcasts the `AllOperatorsInitializedOnNode` message to all other nodes.
4. The node waits until it has received `AllOperatorsInitializedOnNode` messages from all nodes. It then begins executing operators via `Operator::run()`.

In addition to the protocol and control plane, this PR
- Refactors macros for easier maintainability.
- Deprecates the `--addresses` argument.
- Introduces `--data-addresses` and `--control-addresses` arguments.
- Refactors some tests.

### DIFF
--- a/src/communication/mod.rs
+++ b/src/communication/mod.rs
@@ -12,7 +12,7 @@ pub use crate::communication::control_message_codec::ControlMessageCodec;
 pub use crate::communication::endpoints::{RecvEndpoint, SendEndpoint};
 pub use crate::communication::errors::{CodecError, CommunicationError, TryRecvError};
 pub use crate::communication::message_codec::MessageCodec;
-pub use crate::communication::pusher::{ControlPusher, Pusher, PusherT};
+pub use crate::communication::pusher::{Pusher, PusherT};
 pub use crate::communication::serializable::Serializable;
 
 use byteorder::{ByteOrder, NetworkEndian, WriteBytesExt};
@@ -34,7 +34,6 @@ use crate::{dataflow::stream::StreamId, node::node::NodeId, OperatorId};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum ControlMessage {
-    AllOperatorsInitialized,
     AllOperatorsInitializedOnNode(NodeId),
     OperatorInitialized(OperatorId),
     RunOperator(OperatorId),
@@ -65,7 +64,7 @@ impl SerializedMessage {
 }
 
 pub struct ControlMessageHandler {
-    tx: ControlPusher,
+    channels_to_senders: HashMap<NodeId, UnboundedSender<ControlMessage>>,
     rx: UnboundedReceiver<ControlMessage>,
 }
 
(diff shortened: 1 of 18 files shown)


src/communication/receivers.rs
```diff
@@ -86,3 +93,53 @@ pub async fn run_receivers(mut receivers: Vec<ERDOSReceiver>) -> Result<(), Comm
...
+            stream,
+            channel_to_handler,
+        }
+    }
+
+    pub async fn run(&mut self) -> Result<(), CommunicationError> {
+        while let Some(res) = self.stream.next().await {
+            match res {
+                Ok(msg) => {
+                    self.channel_to_handler
```
ICGog: Should we update channel_to_handler regularly to ensure that we have up-to-date mappings between channels and handlers?

src/communication/mod.rs
```diff
@@ -8,35 +8,41 @@ pub mod senders;
...
+    prelude::*,
+    sync::mpsc::{UnboundedReceiver, UnboundedSender},
+};
 
-use crate::node::node::NodeId;
+use crate::{dataflow::stream::StreamId, node::node::NodeId, OperatorId};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum ControlMessage {
     AllOperatorsInitialized,
```
ICGog: Is AllOperatorsInitialized used anywhere? If not, please remove.

src/node/node.rs
```diff
@@ -80,8 +93,7 @@ impl Node {
             let (split_sink, split_stream) = framed.split();
             // Create an ERDOS receiver for the stream half.
             stream_halves.push(
-                ERDOSReceiver::new(node_id, split_stream, self.channels_to_receivers.clone())
-                    .await,
+                ERDOSReceiver::new(node_id, split_stream, self.channels_to_receivers.clone()).await,
```
ICGog: Please fix indentation.

src/node/node.rs
```diff
@@ -92,7 +104,46 @@ impl Node {
...
+        &self,
+        streams: Vec<(NodeId, TcpStream)>,
+    ) -> (
+        ControlMessageHandler,
+        Vec<ControlSender>,
+        Vec<ControlReceiver>,
+    ) {
+        let mut control_receivers = Vec::new();
+        let mut control_senders = Vec::new();
+        let (handler_tx, handler_rx) = mpsc::unbounded_channel();
```
ICGog: What is the rationale of using a `std::mpsc` and not a `tokio::mpsc`? `recv` may block the runtime thread while waiting for a message on the channel.

src/node/node.rs
```diff
@@ -104,22 +155,62 @@ impl Node {
...
         // Execute operators scheduled on the current node.
         let channel_manager = Arc::new(std::sync::Mutex::new(channel_manager));
-        for operator_info in graph
+        let local_operators: Vec<_> = graph
             .get_operators()
             .into_iter()
             .filter(|op| op.node_id == self.id)
-        {
+            .collect();
+        let num_local_operators = local_operators.len();
```
ICGog: Please move closer to where it is used.

src/node/node.rs
```diff
@@ -104,22 +155,62 @@ impl Node {
...
+
+        let (operator_tx, rx_from_operators) = std::sync::mpsc::channel();
```
ICGog: As above, why are we not using tokio's mpsc?

src/node/node.rs
```diff
@@ -104,22 +155,62 @@ impl Node {
...
+            let (tx, rx) = std::sync::mpsc::channel();
+            channels_to_operators.insert(operator_info.id, tx);
             // Launch the operator as a separate async task.
             tokio::spawn(async move {
-                (operator_info.runner)(channel_manager_copy).execute();
+                (operator_info.runner)(channel_manager_copy, operator_tx_copy, rx).execute();
             });
         }
 
+        // Wait for all operators to finish setting up
```
ICGog: Can you please move the code related to waiting for all operators to initialize, and the code broadcasting to all nodes to separate methods?

src/node/node.rs
```diff
@@ -161,6 +269,15 @@ impl Node {
         if let Err(err) = receivers_res {
             error!(self.config.logger, "Error with ERDOS receivers: {:?}", err);
```
ICGog: Please update log message Error with data receives....

PR #63, opened 2020-01-30
Bug fix in run_async python interface.

ICGog:
The additional internal operator causes the ingest_extract integration test to fail with a panic.


pschafhalter:
I'm not sure I agree with this -- if there is 1 operator and an async driver, then there should be 2 nodes running independently (1 for the driver script, 1 for the operator). It looks like there would only be 1 node after these changes.

PR #71, opened 2020-02-24
Implement exporting dataflow graph to DOT file

pschafhalter:
Introduces an API to export the dataflow graph to a [DOT file](https://en.wikipedia.org/wiki/DOT_(graph_description_language)).

In Rust CLI, do `-g=dataflow_graph.dot` or `--graph-filename=dataflow_graph.dot`.

In Python, do `erdos.run(driver=driver, graph_filename="dataflow_graph.dot")`.

In Rust, labels operators with provided names or falls back to operator ID. In python, labels operators with class names. Streams are labeled with Stream IDs.

Example for `python/examples/simple_pipeline.py`:
simple_pipeline


ICGog:
LGTM, but please use an Option for the name instead of a string.

PR #80, opened 2020-02-28
Added decorator for profiling operator methods.

ICGog:


python/erdos/__init__.py
```diff
@@ -189,3 +189,29 @@ def _flow_watermark_callback(timestamp, *write_streams):
...
+                timestamp = args[1]
+            elif isinstance(args[1], Message):
+                # The func is a callback.
+                timestamp = args[1].timestamp
+            else:
+                # The func is a regular method.
+                timestamp = None
+            event_name = op_name + "." + cb_name + "@" + str(timestamp)
+        else:
+            raise TypeError('profile can also be used on operator methods')
```
pschafhalter: Can this use double quotes please?

pschafhalter: and maybe improve the error message?

pschafhalter: And maybe improve the error message?

PR #95, opened 2020-03-18
Add time-versioned state to enforce determinism

pschafhalter:
Adds time-versioned state which ensures that an operator behaves deterministically while allowing as much parallelism as possible.

Time-versioned state enforces 3 different access patterns:
1. When created in Operator::new(). This allows setting the number of past states accessible via the history size and an initial state associated with `Timestamp::bottom()`.
2. From a regular, non-watermark callback. This allows appending messages which are later exposed to watermark callbacks. Appended messages may be compressed versions of ERDOS messages.
3. From a watermark callback. This allows reading appended messages and reading state up until current timestamp. In addition, it allows mutating the state associated with the current timestamp.

For each access pattern, access rules are enforced via the `AccessContext`. ERDOS manages transitions between `AccessContext`s.

Time-versioned state currently has the following weaknesses:
1. Manual GC. Currently, users must manually do `time_versioned_state.close_time(t)` to GC message and state history. This potentially introduces non-determinism in states shared across multiple streams. Fix this by pushing GC to ERDOS; when the last watermark callback for the state at time `t` runs, we can close time for `t`. Unfortunately, this requires logic not currently implemented in ERDOS.
   1. Make an issue for this after merging.
2. Ordering of messages in vector. Watermark callbacks that behave differently depending on the ordering of appended messages will behave non-deterministically. This can be solved by using an unordered collection or sorting the vector before exposing to the user, but these may impact performance or adds more restrictions on what can be appended (e.g. a HashMap with a count requires that the Hash trait is implemented).

### DIFF
--- a/src/dataflow/message.rs
+++ b/src/dataflow/message.rs
@@ -108,6 +108,13 @@ impl IntTimestamp {
         }
     }
 
+    pub fn bottom() -> Self {
+        Self {
+            time: Vec::new(),
+            is_top: false,
+        }
+    }
+
     pub fn is_top(&self) -> bool {
         self.is_top
     }
--- a/src/dataflow/state.rs
+++ b/src/dataflow/state.rs
@@ -69,6 +69,8 @@ pub struct TimeVersionedState<S: State + Default, T: Clone> {
     history_size: usize,
     // Determines access control rules.
     access_context: AccessContext,
+    // TODO: consider replacing the Vec with some unordered data structure as the ordering of messages
+    // leaks information that may break determinism.
     message_history: BTreeMap<Timestamp, Vec<T>>,
     state_history: BTreeMap<Timestamp, S>,
 }
--- a/src/dataflow/state.rs
+++ b/src/dataflow/state.rs
@@ -38,13 +38,15 @@ pub(crate) trait ManagedState {
     fn set_access_context(&mut self, access_context: AccessContext);
     fn set_current_time(&mut self, t: Timestamp);
     /// Garbage collects any state no longer needed up until time t.
-    fn close_time(&mut self, t: &Timestamp);
+    fn close_time(&mut self, t: &Timestamp) -> Result<(), AccessError>;
 }
 
 impl<S: State> ManagedState for S {
     default fn set_access_context(&mut self, _access_context: AccessContext) {}
     default fn set_current_time(&mut self, _t: Timestamp) {}
-    default fn close_time(&mut self, _t: &Timestamp) {}
+    default fn close_time(&mut self, _t: &Timestamp) -> Result<(), AccessError> {
+        Ok(())
+    }
 }
 
 /// Ensures that an operator behaves deterministically while allowing as much
@@ -373,8 +375,8 @@ impl<S: State + Default, T: Clone> ManagedState for TimeVersionedState<S, T> {
             .or_default();
     }
 
-    fn close_time(&mut self, t: &Timestamp) {
-        self.close_time(t);
+    fn close_time(&mut self, t: &Timestamp) -> Result<(), AccessError> {
+        self.close_time(t)
     }
 }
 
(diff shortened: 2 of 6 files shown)


src/dataflow/stream/internal_stateful_read_stream.rs
```diff
@@ -76,17 +80,19 @@ impl<D: Data, S: State> EventMakerT for InternalStatefulReadStream<D, S> {
...
                             for callback in stateful_cbs {
                                 let msg_copy = msg.clone();
-                                (callback)(msg_copy.timestamp, msg_copy.data, unsafe {
-                                    Arc::get_mut_unchecked(&mut state_ref)
-                                })
+                                let state_ref_mut =
+                                    unsafe { Arc::get_mut_unchecked(&mut state_arc) };
+                                state_ref_mut.set_access_context(AccessContext::Callback);
+                                state_ref_mut.set_current_time(msg.timestamp.clone());
+                                (callback)(msg_copy.timestamp, msg_copy.data, state_ref_mut)
```
sukritkalra: Is this safe in a multi-threaded operator? Since multiple mutable references to `self.state` are allowed and there is no locking in the state data structure, is it possible that we set the AccessContext as Callback and start a message callback for time `t + 1`, meanwhile, also start a watermark callback for time `t`, thus changing the AccessContext to WatermarkCallback and allow access to state in the message callback? The code in the operator to achieve this might be a bit contrived, but I think it's possible to do this.

sukritkalra: Similarly, the `current_time` might jump around in a multi-threaded operator.

src/dataflow/message.rs
```diff
@@ -108,6 +108,13 @@ impl IntTimestamp {
         }
     }
 
+    pub fn bottom() -> Self {
+        Self {
+            time: Vec::new(),
+            is_top: false,
+        }
+    }
```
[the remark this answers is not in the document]
pschafhalter: I'm open to that, but maybe let's do that in a follow-up PR? @ICGog what's your view on this?

src/dataflow/state.rs
```diff
@@ -0,0 +1,549 @@
...
+/// ERDOS manages transitions between `AccessContext`s.
+#[derive(Clone)]
+pub struct TimeVersionedState<S: State + Default, T: Clone> {
+    current_time: Timestamp,
+    // The number of past states to keep.
+    history_size: usize,
+    // Determines access control rules.
+    access_context: AccessContext,
+    // TODO: consider replacing the Vec with some unordered data structure as the ordering of messages
+    // leaks information that may break determinism.
```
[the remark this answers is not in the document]
pschafhalter: Sets don't allow for duplicates, but we could move to a HashMap which stores a count.
A HashMap that maps message to count could work, but requires that the messages implement Hash.
I'm not really sure if using a HashMap, or keeping the Vec and adding documentation explaining that there's no guarantees about the order of the messages is the best solution. What do you think?

src/dataflow/stream/internal_stateful_read_stream.rs
pschafhalter: Hmm, I thought that for now only 1 callback can access state at a time. I agree about that scenario for the future --  a better solution might be to clone (could be expensive) or provide a view on the datastructure.

src/dataflow/state.rs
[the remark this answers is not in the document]
ICGog: I'm leaning on the vec side for now. Sure, users can introduce non-determinism, but operator developers can also be careful how they work with the vectors. IMO we can revisit this when we'll have another pass over the API.

src/dataflow/stream/internal_stateful_read_stream.rs
ICGog: Can you please open an issue to track this?

PR #101, opened 2020-03-24
[Operators] Replace CHashMap with locked HashMap

### LINKED ISSUE #100: [Operators] JoinOperator non-deterministically fails with '`CHashMap` scan_mut failed! No entry found.'
There appears to be a [race condition/concurrency bug](https://gitlab.redox-os.org/redox-os/chashmap/issues/3) in chashmap on which the JoinOperator depends.

Fix this by falling back to a mutexed HashMap.

pschafhalter:
Also did some cleanup to reduce build warnings.

Fixes #100


pschafhalter:
It didn't when I ran the tests :(

PR #103, opened 2020-03-25
Call Operator::run from OperatorExecutor to remove busy-looping

### LINKED ISSUE #102: [Rust] Number of operators per node is limited by Tokio thread pool size
The busy looping below causes resource starvation. This causes observable issues when the number operators increases while the number of threads remains constant (I observed it with 4 operators and 4 threads on a node).
https://github.com/erdos-project/erdos/blob/9dcde096c77a48c144deb3455ca4165e4739d8c6/src/lib.rs#L157-L164

pschafhalter:
OperatorExecutors now call `Operator::run` from an async context. This allows waiting on the `RunOperator` signal using a blocking call instead of busy-looping.

Also fixes some bugs in the tests.

Fixes #102.

### DIFF
--- a/src/node/operator_executor.rs
+++ b/src/node/operator_executor.rs
@@ -183,8 +183,12 @@ impl OperatorExecutor {
             .all(|x| x.load(Ordering::SeqCst))
     }
 
-    /// Runs [`OperatorExecutor::execute`] upon receiving [`ControlMessage::RunOperator`] matching this operator's ID.
-    pub async fn execute_on_control_message(&mut self) {
+    /// A high-level execute function that first waits for a [`ControlMessage::RunOperator`] message
+    /// and executes [`Operator::run`].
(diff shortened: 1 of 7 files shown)


src/node/operator_executor.rs
```diff
@@ -179,10 +183,25 @@ impl OperatorExecutor {
...
+    }
+
     /// A high-level execute function that retrieves events from the input streams, adds them to
     /// the lattice being maintained by the executor and notifies the `event_runner` invocations to
     /// process the received events.
-    pub async fn execute(&mut self) {
+    async fn execute(&mut self) {
+        // Callbacks are not invoked while the operator is running.
+        self.operator.run();
+
```
sukritkalra: Can you move the `execute_on_control_message` logic inside `execute`?

PR #106, opened 2020-03-26
[Rust] Add latency benchmarks

pschafhalter:
Adds latency benchmarks using criterion.

Also adds `erdos::reset` which resets the internal dataflow graph, allowing a program to stop a dataflow, reset the graph, and create new dataflow graph.

Contains some minor bugfixes as well.

In addition, ran the benchmarks on erdos4:
[report.zip](https://github.com/erdos-project/erdos/files/4384881/report.zip)


src/lib.rs
```diff
@@ -406,6 +406,15 @@ impl fmt::Display for Uuid {
     }
 }
 
+/// Resets the internal dataflow graph.
```
ICGog: Can you please add a more descriptive comment to this method? (e.g., resets seed and creates a new graph...)

benches/latency.rs
```diff
@@ -0,0 +1,432 @@
...
+
+    let idx = if p >= 1.0 {
+        durations.len() - 1
+    } else {
+        (durations.len() as f32 * p).floor() as usize
+    };
+    durations[idx]
+}
+
+/// Varies the size of a message and measures the resulting latency.
```
ICGog: Please specify in the comment that this benchmark tests 1 to 1.

PR #125, opened 2020-06-23
[Rust] Improve documentation

pschafhalter:
- Improves existing documentation across the crate.
- Hides documentation for internal data structures that are public by necessity due to macros (e.g. `InternalReadStream`).
- Moves macros to a separate `connect.rs` submodule to clean up `lib.rs`.
- Adds building documentation to CI. This is in a separate workflow file so that we have the option to host up-to-date documentation for the current master in the future.

You can view the updated documentation [here](https://people.eecs.berkeley.edu/~pschafhalter/pub/erdos/doc/erdos/) (I might take this offline after the PR gets merged).


src/lib.rs
```diff
@@ -1,28 +1,137 @@
...
 //! `ERDOS` is designed for low latency. Self-driving car pipelines require
 //! end-to-end deadlines on the order of hundreds of milliseconds for safe
 //! driving. Similarly, self-driving cars typically process gigabytes per
 //! second of data on small clusters. Therefore, `ERDOS` is optimized to
 //! send small amounts of data (gigabytes as opposed to terabytes)
 //! as quickly as possible.
 //!
-//! `ERDOS` provides determinisim through **watermarks**. Low watermarks
+//! ## Determinism
+//! ERDOS provides determinisim through **watermarks**. Low watermarks
```
ICGog, line 134: Can you please also add a section on watermarks because they're a core feature of ERDOS? I think we also need to explain top & bottom watermarks.

PR #135, opened 2020-07-26
Refactor operator event ordering and lattice

pschafhalter:
- Changes operator event ordering s.t. `x < y` implies `x` *precedes* `y`.
- Reverse edges on lattice dependency graph to match [convention](https://en.wikipedia.org/wiki/Dependency_graph).
- Add warning if more than 25 events queued in lattice.
- Add `to_dot` method to lattice for easier visualization/debugging.
- Code refactoring and comments on lattice.
- Tests use `assert` instead of `assert_eq(*, true)`.

### DIFF
--- a/src/node/lattice.rs
+++ b/src/node/lattice.rs
@@ -8,7 +8,7 @@ use std::{
 use futures::lock::Mutex;
 use petgraph::{
     dot::{self, Dot},
-    stable_graph::{EdgeIndex, NodeIndex, StableGraph},
+    stable_graph::{NodeIndex, StableGraph},
     visit::{DfsPostOrder, Reversed},
     Direction,
 };
@@ -119,7 +119,7 @@ impl PartialOrd for RunnableEvent {
 ///
 /// Events can be added to the lattice using the `add_events` function, and retrieved using the
 /// `get_event` function. The lattice requires a notification of the completion of the event using
-/// the `mark_as_compeleted` function in order to unblock dependent events, and make them runnable.
+/// the `mark_as_completed` function in order to unblock dependent events, and make them runnable.
 ///
 /// # Example
 /// The below example shows how to insert events into the Lattice and retrieve runnable events from
@@ -368,34 +368,28 @@ impl ExecutionLattice {
             .expect("Item must be in the leaves of the lattice.");
         leaves.remove(event_idx);
 
-        // Go over the parents of the node, and check which ones have no dependencies on other
-        // nodes, and add them to the list of the leaves.
-        let mut edges_to_remove: Vec<NodeIndex<u32>> = Vec::new();
-        for parent_id in forest.neighbors_directed(node_idx, Direction::Incoming) {
-            // Promote parent to leaf if it does not depend on any other events.
+        // Capture the parents of the node.
+        let parent_ids: Vec<NodeIndex> = forest
+            .neighbors_directed(node_idx, Direction::Incoming)
+            .collect();
+
+        // Remove the node from the graph. This will also remove edges from the parents.
+        forest.remove_node(node_idx);
+
+        // Promote parents to leaves if they have no dependencies, and add their corresponding
+        // events to the run queue.
+        for parent_id in parent_ids {
             if forest
                 .neighbors_directed(parent_id, Direction::Outgoing)
                 .count()
-                == 1
+                == 0
             {
                 let timestamp: Timestamp = forest[parent_id].as_ref().unwrap().timestamp.clone();
                 let parent = RunnableEvent::new(parent_id).with_timestamp(timestamp);
                 leaves.push(parent.clone());
                 run_queue.push(parent);
             }
-
-            // Remove the edge from the parent.
-            edges_to_remove.push(parent_id);
         }
-
-        // Remove the edges from the forest.
-        for parent_id in edges_to_remove {
-            let edge_id: EdgeIndex<u32> = forest.find_edge(parent_id, node_idx).unwrap();
-            forest.remove_edge(edge_id);
-        }
-
-        // Remove the completed event from the forest.
-        forest.remove_node(node_idx);
     }
 
     /// Convert graph to string in DOT format.
--- a/src/node/lattice.rs
+++ b/src/node/lattice.rs
@@ -811,56 +811,106 @@ mod test {
         );
     }
 
-    /// Test that the messages are retrieved in the timestamp order.
-    /// Added to address the bug where we would have starvation due to
-    /// watermark messages of earlier timestamps being blocked by non-watermark
-    /// callbacks of newer timestamps.
+    /// Tests that duplicate events do not end up in the lattice's leaves or
+    /// run queue. This can happen if duplicate edges exist in the dependency
+    /// graph.
     #[test]
-    fn test_ordered_concurrent_execution() {
-        let lattice: ExecutionLattice = ExecutionLattice::new();
-        let events = vec![
+    fn test_no_duplicates() {
+        let lattice = ExecutionLattice::new();
+        // Add 2 operators that can run concurrently.
+        let initial_events = vec![
             OperatorEvent::new(
-                Timestamp::new(vec![2]),
+                Timestamp::new(vec![0]),
                 false,
                 0,
                 HashSet::new(),
                 HashSet::new(),
-                || (),
+                || {},
             ),
             OperatorEvent::new(
-                Timestamp::new(vec![1]),
-                true,
-                0,
-                HashSet::new(),
-                HashSet::new(),
-                || (),
-            ),
-            OperatorEvent::new(
-                Timestamp::new(vec![3]),
+                Timestamp::new(vec![0]),
                 false,
                 0,
                 HashSet::new(),
                 HashSet::new(),
-                || (),
+                || {},
             ),
         ];
-        block_on(lattice.add_events(events));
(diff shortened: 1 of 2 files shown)


src/node/lattice.rs
```diff
@@ -188,90 +191,81 @@ impl ExecutionLattice {
...
+                // Begin a DFS at the specified root.
+                dfs.move_to(root.node_index);
+                while let Some(nx) = dfs.next(Reversed(&*forest)) {
+                    // If any of the current node's children precede the added event, then the current node also precedes
+                    // the added event. By induction, the root would also precede the added event, so we can move on to the
+                    // next root.
+                    for child in forest.neighbors_directed(nx, Direction::Incoming) {
+                        if preceding_events.contains(&child) {
+                            preceding_events.insert(nx);
+                            continue 'dfs_roots;
```
pschafhalter, line 216: Found [this Rust syntax](https://doc.rust-lang.org/reference/expressions/loop-expr.html#continue-expressions) surprisingly elegant.

src/node/lattice.rs
```diff
@@ -298,30 +290,32 @@ impl ExecutionLattice {
...
+            if preceding_events.is_empty() {
                 roots.push(RunnableEvent::new(event_ix).with_timestamp(event_timestamp.clone()));
                 run_queue.push(RunnableEvent::new(event_ix).with_timestamp(event_timestamp));
             }
         }
+
+        if forest.node_count() > 25 {
+            slog::warn!(
+                crate::get_terminal_logger(),
+                "{} operator events queued in lattice. Increase number of operator executors or decrease incoming \
```
ICGog, line 315: IMO, it would be useful to also include the operator name/id in this log message. We can do this in a different change because we'll also have to pass the id/name to the lattice.

src/node/lattice.rs
```diff
@@ -188,90 +191,81 @@ impl ExecutionLattice {
...
-                    // Check if any of the children of the current node in DFS traversal already have a
-                    // dependency on the node we are trying to add.
-                    let mut dependent_child: bool = false;
-                    for child in forest.neighbors(nx) {
-                        if dependencies.contains(&child) {
-                            dependencies.insert(nx);
-                            dependent_child = true;
-                            break;
+            // Visits each node in the graph once.
+            let mut dfs = DfsPostOrder::empty(Reversed(&*forest));
```
ICGog, line 195: Do we know what's the cost of Reversed? Does it require a pass over the forest?

ICGog:
@sukritkalra can you please also have a look over this PR?

src/node/lattice.rs
```diff
@@ -188,90 +191,81 @@ impl ExecutionLattice {
...
-                while let Some(nx) = dfs.next(&*forest) {
-                    // Check if any of the children of the current node in DFS traversal already have a
-                    // dependency on the node we are trying to add.
-                    let mut dependent_child: bool = false;
-                    for child in forest.neighbors(nx) {
-                        if dependencies.contains(&child) {
-                            dependencies.insert(nx);
-                            dependent_child = true;
-                            break;
+            // Visits each node in the graph once.
```
sukritkalra, line 194: Can we add a comment here about why we went through a reversal of the forest instead of the way it was originally?

src/node/lattice.rs
```diff
@@ -157,7 +159,8 @@ impl PartialOrd for RunnableEvent {
 /// ```
 pub struct ExecutionLattice {
     /// The `forest` is the directed acyclic graph that maintains the dependency graph of the
-    /// events.
+    /// events. The relation A -> B means that A *depends on* B.This dependency also indicates that
+    /// B precedes A (B < A) in the ordering. An event can be executed if it has no outbound edges.
     forest: Arc<Mutex<StableGraph<Option<OperatorEvent>, ()>>>,
```
sukritkalra, line 164: Can we actually skip the whole `Reversed` thing on each `add_events` by just changing this to
`forest: Arc<Mutex<Reversed<StableGraph<Option<OperatorEvent>, ()>>>>`?

src/node/lattice.rs
pschafhalter: Logically, it's the same as before, but instead of keeping a boolean to continue the outer loop, the Rust syntax lets us continue the outer loop directly :)

src/node/lattice.rs
```diff
@@ -184,94 +187,93 @@ impl ExecutionLattice {
...
+                                Vec::new();
+                            for child in forest.neighbors_directed(nx, Direction::Incoming) {
                                 let child_node: &OperatorEvent =
                                     forest.node_weight(child).unwrap().as_ref().unwrap();
-                                if child_node < &event {
-                                    // The child has a dependency on the node being added.
+                                if child_node > &event {
+                                    // The child depends on the added event.
                                     // Break the dependency from the DFS node to its child, and add a
                                     // dependency from the node to be added to the child.
```
sukritkalra, line 241: Fix comment.

src/node/lattice.rs
```diff
@@ -353,52 +355,74 @@ impl ExecutionLattice {
...
         // Go over the children of the node, and check which ones have no dependencies on other
-        // nodes, and add them to the list of the roots.
-        let mut remove_edges: Vec<NodeIndex<u32>> = Vec::new();
-        let mut root_nodes: Vec<RunnableEvent> = Vec::new();
-        for child_id in forest.neighbors(node_id) {
-            // Does this node have any other incoming edge?
+        // nodes, and add them to the list of the leaves.
+        let mut edges_to_remove: Vec<NodeIndex<u32>> = Vec::new();
+        for child_id in forest.neighbors_directed(node_idx, Direction::Incoming) {
+            // Promote child to root if it does not depend on any other events.
```
sukritkalra, line 375: Fix comment.

src/node/lattice.rs
```diff
@@ -184,94 +187,93 @@ impl ExecutionLattice {
...
                                 }
                             }
                         }
                     } else {
                         // Reached a node that is already executing, but hasn't been completed.
-                        // The current node will probably get added as a root. Add dependencies to the
-                        // children, if any.
-                        for child in forest.neighbors(nx) {
+                        // The current node will probably get added as a root. Add dependencies
+                        // between children and current event.
```
sukritkalra, line 272: Fix comment.

src/node/lattice.rs
```diff
@@ -184,94 +187,93 @@ impl ExecutionLattice {
...
+                        for child in forest.neighbors_directed(nx, Direction::Incoming) {
```
sukritkalra, line 273: `parent`

src/node/lattice.rs
```diff
@@ -184,94 +187,93 @@ impl ExecutionLattice {
...
                             let child_node = forest.node_weight(child).unwrap().as_ref().unwrap();
```
sukritkalra, line 274: `parent_node`

src/node/lattice.rs
```diff
@@ -282,46 +284,46 @@ impl ExecutionLattice {
...
                 for item in old_run_queue {
-                    if !remove_roots.contains(&item) {
+                    if !demoted_leaves.contains(&item) {
                         run_queue.push(item);
                     }
                 }
             }
 
-            // If no dependencies of this event, then we can safely create a new root in the forest and
+            // If the added event depends on no others, then we can safely create a new root in the forest and
```
sukritkalra, line 312: Fix comment.

src/node/lattice.rs
```diff
@@ -353,52 +355,74 @@ impl ExecutionLattice {
...
 
         // Go over the children of the node, and check which ones have no dependencies on other
-        // nodes, and add them to the list of the roots.
-        let mut remove_edges: Vec<NodeIndex<u32>> = Vec::new();
-        let mut root_nodes: Vec<RunnableEvent> = Vec::new();
-        for child_id in forest.neighbors(node_id) {
-            // Does this node have any other incoming edge?
+        // nodes, and add them to the list of the leaves.
+        let mut edges_to_remove: Vec<NodeIndex<u32>> = Vec::new();
+        for child_id in forest.neighbors_directed(node_idx, Direction::Incoming) {
```
sukritkalra, line 374: `parent_id`

src/node/lattice.rs
```diff
@@ -353,52 +355,74 @@ impl ExecutionLattice {
...
+        let event = RunnableEvent::new(node_idx);
+        let event_idx = leaves
             .iter()
             .position(|e| e == &event)
-            .expect("Item must be in the roots of the lattice.");
-        roots.remove(event_idx);
+            .expect("Item must be in the leaves of the lattice.");
+        leaves.remove(event_idx);
 
         // Go over the children of the node, and check which ones have no dependencies on other
```
sukritkalra, line 371: Fix comment.

src/node/lattice.rs
```diff
@@ -353,52 +355,74 @@ impl ExecutionLattice {
...
+            // Promote child to root if it does not depend on any other events.
             if forest
-                .neighbors_directed(child_id, Direction::Incoming)
+                .neighbors_directed(child_id, Direction::Outgoing)
                 .count()
                 == 1
             {
                 let timestamp: Timestamp = forest[child_id].as_ref().unwrap().timestamp.clone();
-                root_nodes.push(RunnableEvent::new(child_id).with_timestamp(timestamp));
+                let child = RunnableEvent::new(child_id).with_timestamp(timestamp);
```
sukritkalra, line 382: `parent`

PR #136, opened 2020-08-05
Logging and Documentation for Python API of ERDOS.

sukritkalra:
- Introduces a `logger` in the top-level ERDOS module that users can use to enable more verbose logging for Python.
- Improves documentation for front-facing Python classes and functions.
- Adds type hints wherever possible (unsure about the TODOs since the API does not seem to be well-formed there)
- Introduces a `name` property in all the Streams (default: `None`) that future refactorings of the Rust codebase will automatically populate for the Python streams (for now).


python/erdos/streams.py
```diff
@@ -31,24 +37,47 @@ def _to_py_message(msg):
...
+    def name(self) -> Union[str, None]:
+        """ The name of the stream. (`None` if no name was given.) """
+        return self._name
+
+    def is_closed(self) -> bool:
         """Whether a top watermark message has been received."""
         return self._py_read_stream.is_closed()
 
-    def read(self):
+    # TODO (Sukrit) :: What is the return type of `read`?
```
ICGog, line 79: In the current implementation it's Message.

python/erdos/streams.py
```diff
@@ -135,54 +196,100 @@ class LoopStream(object):
...
+    """An :py:class:`IngestStream` enables drivers to inject data into a
+    running ERDOS application.
+    
+    The driver can initialize a new :py:class:`IngestStream` and connect it to
+    an :py:class:`Operator` through :py:func:`connect`. Similar to a 
+    :py:class:`WriteStream`, an :py:class:`IngestStream` provides a 
+    :py:func:`IngestStream.send` to enable the driver to send data to the 
+    operator to which it was connected.
+    """
+    def __init__(self, _name: Union[str, None] = None):
```
ICGog, line 224: Can you please add the name type annotation to the other constructors as well (e.g., WriteStream)?

python/erdos/__init__.py
```diff
@@ -61,85 +95,98 @@ def connect(op_type, config, read_streams, *args, **kwargs):
...
+        use.
     """
+    logger.info("Resetting the default graph.")
     global _num_py_operators
     _num_py_operators = 0
     _internal.reset()
 
 
-def run(graph_filename=None, start_port=9000):
+# TODO (Sukrit) : Should this be called a GraphHandle?
```
ICGog, line 123: I think we might have to do a bit of work to correctly shutdown the dataflow. AFAIk, the NodeHandle is only terminating the processes running on the current machine.

@pschafhalter Do you agree? Should we open an issue about this so that we don't forget about this issue.

python/erdos/__init__.py
```diff
@@ -61,85 +95,98 @@ def connect(op_type, config, read_streams, *args, **kwargs):
...
+def run_async(graph_filename: Optional[str] = None,
+              start_port: Optional[int] = 9000) -> NodeHandle:
     """Instantiates and runs the dataflow graph asynchronously.
 
     ERDOS will spawn 1 process for each python operator, and connect them via
     TCP.
 
     Args:
-        graph_filename (str): the filename to which to write the dataflow graph
+        graph_filename: The filename to which to write the dataflow graph
```
ICGog, line 181: What guide lines are you following? Are we removing the types from the comments if we're using type annotations?

python/erdos/__init__.py
pschafhalter: Yes, let's make an issue. Although currently we don't expose a way to distribute pipelines across multiple machines when using Python, so in a way this isn't a problem right now.

python/erdos/__init__.py
```diff
@@ -177,28 +226,41 @@ def sigint_handler(sig, frame):
...
-        write_streams (list of WriteStream): streams on which the callback
-            can send messages.
-        callback (timestamp, list of WriteStream -> None): a low watermark
-            callback.
+        read_streams: Streams on which the callback is invoked.
+        write_streams: Streams on which the callback can send messages.
+        callback: The callback to be invoked upon receipt of a 
+            :py:class:`.WatermarkMessage` on all the `read_streams`.
     """
+    logger.debug("Add watermark callback {name} to the input streams: {_input}"
```
pschafhalter, line 240: Can you change this to "Adding" to keep the same tense across logger messages?

python/erdos/streams.py
```diff
@@ -62,35 +91,51 @@ def try_read(self):
...
-            write_streams (list of WriteStream): write streams passed to the
-                callback.
+            callback: A callback that takes a message and a sequence of 
+                :py:class:`WriteStream` s.
+            write_streams: Write streams passed to the callback.
         """
         if write_streams is None:
             write_streams = []
 
+        logger.debug("Add callback {name} to the input stream {_input}, "
```
pschafhalter, line 107: Adding

python/erdos/streams.py
```diff
@@ -62,35 +91,51 @@ def try_read(self):
...
-            write_streams (list of WriteStream): write streams passed to the
-                callback.
+            callback: A callback that takes a message and a sequence of 
+                :py:class:`WriteStream` s.
+            write_streams: Write streams passed to the callback.
         """
         if write_streams is None:
             write_streams = []
+        logger.debug(
+            "Add watermark callback {name} to the input stream "
```
pschafhalter, line 134: Adding

python/erdos/streams.py
```diff
@@ -101,32 +146,48 @@ def internal_watermark_callback(coordinates, is_top):
...
-
-    Handlese serialization of messages, and wraps an `internal.PyWriteStream`.
+    """ A :py:class:`WriteStream` allows an :py:class:`Operator` to send 
+    messages and watermarks to other operators that connect to the 
+    corresponding :py:class:`ReadStream`.
 
-    `_py_write_stream` is set during erdos.run(), and should never be set
-    manually.
+    Note:
+        `_py_write_stream` is set during erdos.run(), and should never be set
```
pschafhalter, line 154: Should be :pyfunc:`run` ?

python/erdos/streams.py
```diff
@@ -101,32 +146,48 @@ def internal_watermark_callback(coordinates, is_top):
...
+    def __init__(self, _py_write_stream=None, _name=None, _id=None):
         self._py_write_stream = PyWriteStream(
         ) if _py_write_stream is None else _py_write_stream
+        self._name = _name
+        self._id = _id
 
-    def is_closed(self):
+    @property
+    def name(self) -> Union[str, None]:
+        """ The name of the stream. (`None` if no name was given.) """
```
pschafhalter, line 165: Remove parentheses.

python/erdos/streams.py
```diff
@@ -101,32 +146,48 @@ def internal_watermark_callback(coordinates, is_top):
...
         return self._py_write_stream.is_closed()
 
-    def send(self, msg):
+    def send(self, msg: Message):
         """Sends a message on the stream.
 
         Args:
-            msg (Message): the message to send. This may be a `Watermark` or a
-                `Message`.
+            msg: the message to send. This may be a `Watermark` or a `Message`.
```
pschafhalter, line 176: pyclass:`Watermark` and pyclass:`Message`? Might need to specify pyclass:`erdos.Message` if Sphinx isn't detecting it.

python/erdos/streams.py
```diff
@@ -135,54 +196,100 @@ class LoopStream(object):
...
     Must call `set` on a ReadStream to complete the loop.
     """
-    def __init__(self):
+    def __init__(self, _name: str = None):
         self._py_loop_stream = PyLoopStream()
+        self._name = _name
+
+    @property
+    def name(self) -> Union[str, None]:
+        """ The name of the stream. (`None` if no name was given.) """
```
pschafhalter, line 205: Remove parentheses.

python/erdos/streams.py
```diff
@@ -135,54 +196,100 @@ class LoopStream(object):
...
+    :py:func:`IngestStream.send` to enable the driver to send data to the 
+    operator to which it was connected.
+    """
+    def __init__(self, _name: Union[str, None] = None):
+        self._py_ingest_stream = PyIngestStream(0, _name)
+        self._name = _name
+
+    @property
+    def name(self) -> Union[str, None]:
+        """ The name of the stream. (`None` if no name was given.) """
```
pschafhalter, line 230: Remove parentheses.

python/erdos/__init__.py
pschafhalter: [Google style guide gets rid of types in parentheses](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) when there's type hints. After discussing with @sukritkalra, we might want to keep around types in parentheses for attributes because Sphinx can't currently infer those types.

python/erdos/__init__.py
```diff
@@ -61,85 +95,98 @@ def connect(op_type, config, read_streams, *args, **kwargs):
...
+    logger.info("Resetting the default graph.")
     global _num_py_operators
     _num_py_operators = 0
     _internal.reset()
 
 
-def run(graph_filename=None, start_port=9000):
+# TODO (Sukrit) : Should this be called a GraphHandle?
+# What is the significance of the "Node" here?
+class NodeHandle(object):
```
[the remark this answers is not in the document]
sukritkalra: Seems like with type annotations, Python requires the types to be defined first in the file before they can be used.

I had to move this before `run_async` in order to add a return annotation on it.

python/erdos/__init__.py
```diff
@@ -177,28 +226,41 @@ def sigint_handler(sig, frame):
...
 
-    py_read_streams = [s._py_read_stream for s in read_streams]
-    _internal.add_watermark_callback(py_read_streams,
-                                     internal_watermark_callback)
+    _internal.add_watermark_callback(
+        list(map(attrgetter('_py_read_stream'), read_streams)),
+        internal_watermark_callback)
 
 
 def _flow_watermark_callback(timestamp, *write_streams):
```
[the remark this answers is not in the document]
sukritkalra: I might not be in sync with the master. Let me rebase the PR.

python/erdos/operator.py
```diff
@@ -7,20 +7,46 @@
...
+        matching the read streams and write streams in 
+        :py:func:`Operator.connect`.
 
-        Invoked automatically during erdos.run().
+        Invoked automatically during :py:func:`.run`.
 
-        ERDOS operators never need to call super().__init__() because
-        setup is handled by the ERDOS backend code.
+        Note:
+            An ERDOS operator implementation should not call :py:func:`super`
```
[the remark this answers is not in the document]
sukritkalra: Thanks, I missed this!

PR #138, opened 2020-08-07
Documentation and logging for Rust streams.

sukritkalra:
- Added documentation for Rust streams.
- Removed the instances of logger being passed around in OperatorExecutor. Everything uses `crate::TERMINAL_LOGGER`.
- Added a `new_with_name` to all streams to match existing function on IngestStream.
- Cleaned up API and fixed some inconsistencies.


src/dataflow/stream/errors.rs
```diff
@@ -11,6 +11,8 @@ pub enum ReadError {
     Closed,
 }
 
+// TODO (Sukrit) :: Should we deprecate this? We should have a single ReadError that includes
+// Empty.
```
pschafhalter, line 15: Rust's standard library uses a similar pattern when receiving on an `mpsc::channel`
https://doc.rust-lang.org/nightly/std/sync/mpsc/enum.TryRecvError.html

src/dataflow/stream/ingest_stream.rs
```diff
@@ -31,16 +65,41 @@ impl<D> IngestStream<D>
 where
     for<'a> D: Data + Deserialize<'a>,
 {
+    // TODO (Sukrit) :: Why does this constructor require the NodeID? Inconsistency between
+    // constructors of different types of streams.
```
pschafhalter, line 69: `IngestStream` is used to send data from a particular node into the dataflow. Because ERDOS currently requires all nodes to run the same program, we need to specify which node uses this `IngestStream`.

src/dataflow/stream/ingest_stream.rs
```diff
@@ -31,16 +65,41 @@ impl<D> IngestStream<D>
 where
     for<'a> D: Data + Deserialize<'a>,
 {
+    // TODO (Sukrit) :: Why does this constructor require the NodeID? Inconsistency between
```
ICGog, line 68: I suspect this is because one might want to ingest data in parallel across different nodes.

src/dataflow/stream/read_stream.rs
```diff
@@ -9,66 +9,190 @@ use super::{
     IngestStream, InternalReadStream, LoopStream, StatefulReadStream, StreamId, WriteStream,
 };
 
-/// Allows reading from a stream and registering callbacks.
+/// A [`ReadStream`] allows operators to read data from a corresponding [`WriteStream`].
 ///
-/// Passed as an argument to `Operator::new` and `Operator::connect`.
-/// Also used in the driver to construct a dataflow.
+/// An [`Operator`](crate::dataflow::operator::Operator) receives a [`ReadStream`] in its `new` and
```
pschafhalter, line 14: "gains access to" instead of "receives" is a bit clearer.

src/dataflow/stream/read_stream.rs
```diff
@@ -9,66 +9,190 @@ use super::{
...
+/// `connect` functions. Operators can register both message and watermark callbacks on the
```
pschafhalter, line 15: Could you specify that these are registered in `new`, not `connect`?

src/dataflow/stream/read_stream.rs
```diff
@@ -9,66 +9,190 @@ use super::{
...
+    /// Note that ERDOS automatically converts the [`WriteStream`]s returned by an
+    /// [`Operator`](crate::dataflow::operator::Operator) to a corresponding [`ReadStream`].
     pub fn new() -> Self {
         Self {
             internal_stream: Rc::new(RefCell::new(InternalReadStream::new())),
         }
     }
 
-    /// Add a callback to be invoked when the stream receives a message.
+    /// Returns a new instance of the [`ReadStream`] with the given name..
```
pschafhalter, line 102: Can you fix the typo (2 periods)?

PR #147, opened 2020-11-25
[Python] Add erdos_to_ros operator with success and failure tests.

shreyask23:
Created a new tests directory under erdos/python for python tests. Added erdos_to_ros operator with success and failure tests. The operator converts erdos messages to ros messages and publishes them.


python/tests/erdos_to_ros_test.py
```diff
@@ -0,0 +1,173 @@
...
+    Args:
+        publish_stream (:py:class:`erdos.ReadStream`): Stream on which the
+            operator receives control commands.
+        ros_msg_type: Type used to publish to ros messages
+        rostopic: rostopic to publish messages to
+        node_name: Ros node name
+        func: Callback function that converts a single ros_msg to an erdos_msg
+    """
+
+    def __init__(self, publish_stream, ros_msg_type, rostopic, node_name, func):
```
sukritkalra, line 63: Can we name the function to `conversion_func` from `func`?

Also, we should probably move the order of these arguments around.

python/tests/erdos_to_ros_test.py
```diff
@@ -0,0 +1,169 @@
...
+                "Ros publisher was not defined, \
+                possibly because run() was not called: %s", err)
+        except (ROSException, ROSSerializationException) as err:
+            self.logger.debug("The erdos msg couldn't publish to ros: %s", err)
+
+    def run(self):
+        # Initialize a publisher
+        self.ros_pub = rospy.Publisher(self.rostopic,
+                                       self.ros_msg_type,
+                                       queue_size=10)
```
sukritkalra, line 102: The `queue_size` should not be set to a default since its choice has repercussions on the performance.

sukritkalra (changes requested):
@shreyask23 The operators need to be renamed and move to a proper location in the module instead of the tests.

@pschafhalter Can you take a look? and where do you think we should move the operators to?

pschafhalter:
Operators should be move into `python/erdos/operators/`.

I'll take a closer look once @sukritkalra's comments are addressed.

sukritkalra:
Closing because #188 was merged.

PR #150, opened 2021-02-14
Finished adding checkpoint and restore functionality to simple_pipeli…

sguduguntla:
…ne.py


sukritkalra:
Closing because too stale and WIP.

PR #151, opened 2021-02-23
[Python] Add support for multimachine operators.

shreyask23:
Adds support for multimachine operators. Operators can now be configured with IP addresses and ports for the machines on which they should run.

### Changes

* The `OperatorConfig` class in operator.py now takes in an optional `addr` attribute that defines the address of the machine on which the operator should run.
* The `connect` method and `run_async` method in __ init __.py now keep track of the addresses on which any operators run via two new global variables, node_addrs and num_procs. They use these addresses to set up the networking to connect these operators.
* The file multimachine_pipeline.py provides an example use case for multimachine operators.

### DIFF
--- a/python/erdos/__init__.py
+++ b/python/erdos/__init__.py
@@ -18,6 +18,8 @@ from erdos.timestamp import Timestamp
 import erdos.utils
 
 _num_py_operators = 0
+node_addrs = dict()
+num_procs = dict()
 
 # Set the top-level logger for ERDOS logging.
 # Users can change the logging level to the required level by calling setLevel
@@ -83,6 +85,14 @@ def connect(
     node_id = _num_py_operators
     logger.debug("Connecting operator #{num} ({name}) to the graph.".format(
         num=node_id, name=config.name))
+    if config._addr != None:
+        global node_addrs
+        global num_procs
+        if len(node_addrs) == 0:
+            num_procs[config._addr] = 1
+            node_addrs[0] = (config._addr, 1)
+        num_procs[config._addr] = num_procs.get(config._addr, 0) + 1
+        node_addrs[node_id] = (config._addr, num_procs[config._addr])
 
     py_read_streams = []
     op_read_streams = list(
@@ -120,6 +130,10 @@ def reset():
     logger.info("Resetting the default graph.")
     global _num_py_operators
     _num_py_operators = 0
+    global node_addrs
+    node_addrs.clear()
+    global num_procs
+    num_procs.clear()
     _internal.reset()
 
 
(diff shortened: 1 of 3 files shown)


python/erdos/operator.py
```diff
@@ -147,12 +147,14 @@ def __init__(self,
                  flow_watermarks: bool = True,
                  log_file_name: str = None,
                  csv_log_file_name: str = None,
-                 profile_file_name: str = None):
+                 profile_file_name: str = None,
+                 addr: str = None):
```
pschafhalter, line 151: Rename to `address`

python/erdos/operator.py
```diff
@@ -147,12 +147,14 @@ def __init__(self,
...
         self._name = name
         self._flow_watermarks = flow_watermarks
         self._log_file_name = log_file_name
         self._csv_log_file_name = csv_log_file_name
         self._profile_file_name = profile_file_name
+        self._addr = addr
```
pschafhalter, line 157: `self._address`

python/erdos/__init__.py
```diff
@@ -83,6 +85,14 @@ def connect(
     node_id = _num_py_operators
     logger.debug("Connecting operator #{num} ({name}) to the graph.".format(
         num=node_id, name=config.name))
+    if config._addr != None:
```
pschafhalter, line 88:
```suggestion
    if config._addr is not None:
```

python/erdos/__init__.py
```diff
@@ -18,6 +18,8 @@
 import erdos.utils
 
 _num_py_operators = 0
+node_addrs = dict()
```
pschafhalter, line 21: Rename to `node_to_address`.

Could you add a comment that this maps each Node ID to an addresses and a port offset?

python/erdos/__init__.py
```diff
@@ -18,6 +18,8 @@
...
+num_procs = dict()
```
pschafhalter, line 22: Remove this global variable since we can calculate it from the previous dict (see below).

python/erdos/__init__.py
```diff
@@ -83,6 +85,14 @@ def connect(
     node_id = _num_py_operators
     logger.debug("Connecting operator #{num} ({name}) to the graph.".format(
         num=node_id, name=config.name))
+    if config._addr != None:
+        global node_addrs
+        global num_procs
```
pschafhalter, line 90: Rename `num_procs` to `processes_per_address`, and compute as follows:

```suggestion
        addresses = list(node_to_address.values())
        processes_per_address = {a: addresses.count(a) for a in set(addresses)}
```

python/erdos/__init__.py
```diff
@@ -192,24 +206,56 @@ def run_async(graph_filename: Optional[str] = None,
...
+        '''
+        data_addresses = ["127.0.0.1:{port}".format(port=start_port)] + [
+            node_addrs[i][0] + ":{port}".format(port=start_port + node_addrs[i][1])
+            for i in range(1, _num_py_operators + 1)
+        ]
+        control_addresses = ["127.0.0.1:{port}".format(port=start_port + len(data_addresses))] + [
+            node_addrs[i][0] + ":{port}".format(port=start_port + len(data_addresses) + node_addrs[i][1])
+            for i in range(1, _num_py_operators + 1)
+        ]
+        '''
```
pschafhalter, line 231: remove

python/erdos/__init__.py
```diff
@@ -192,24 +206,56 @@ def run_async(graph_filename: Optional[str] = None,
...
-        for i in range(1, _num_py_operators + 1)
-    ]
+    if not multimachine:
+        processes = [
+            mp.Process(target=runner, args=(i, data_addresses, control_addresses))
+            for i in range(1, _num_py_operators + 1)
+        ]
+    else:
+        import socket
+        local_addr = socket.gethostbyname(socket.gethostname())
```
pschafhalter, line 253: local_address

pschafhalter (commented):
Great start! Needs a bit of cleanup but otherwise looks good. I will try this out to see if it works.

python/erdos/__init__.py
```diff
@@ -83,6 +84,17 @@ def connect(
     node_id = _num_py_operators
     logger.debug("Connecting operator #{num} ({name}) to the graph.".format(
         num=node_id, name=config.name))
+    if config._address is not None:
+        global node_to_address
+        addresses = list(node_to_address.values())
+        processes_per_address = {a: addresses.count(a) for a in set(addresses)}
+        if len(node_to_address) == 0:
+            processes_per_address[config._address] = 1
+            node_to_address[0] = (config._address, 1)
```
pschafhalter, line 93: Could you explain why this is needed?

python/erdos/__init__.py
```diff
@@ -83,6 +84,17 @@ def connect(
...
+        processes_per_address[config._address] = \
```
pschafhalter, line 94: Could you wrap the assignment in parenthesis rather than using `\`?

E.g. 
```python
processes_per_address[config._address] = (
            processes_per_address.get(config._address, 0) + 1)
```

python/erdos/__init__.py
```diff
@@ -83,6 +84,17 @@ def connect(
...
+            processes_per_address.get(config._address, 0) + 1
+        node_to_address[node_id] = (config._address,
+                                    processes_per_address[config._address])
```
pschafhalter, line 97: I'm not sure about this entire section -- I think it could be done in a clearer way. See my review comment.

pschafhalter (changes requested):
Getting close, but could you make the following change to improve clarity?

Thinking about this further, it seems a bit counter-intuitive to store both the address and a port offset in `node_to_address`. I think it would be better to just store the node -> address mapping and directly compute the port for each node in `run_async`.
This would also eliminate the need for `processes_per_address` in `connect`.

PR #152, opened 2021-03-11
[Bugfix] Destroy operator when run finishes

pschafhalter:
Fixes a bug in operators with no input streams where `destroy()` would not be invoked after `run()` finishes.

Note: although `destroy()` will be invoked, top watermarks are *not* automatically sent on output streams.


ICGog (approved):
LGTM

PR #153, opened 2021-03-22
Introduce MessageProcessorT and OperatorExecutorHelper abstractions.

sukritkalra:
PR to track the deduplication of the `OperatorExecutor` code. Changes are as follows:

- Require the operator traits to be both `Send` and `Sync`.
- Introduce `MessageProcessorT` trait that must be implemented by each executor and provides a way to create new `OperatorEvent`s for stateless and stateful message callbacks along with watermark callbacks.
- Introduce `OperatorExecutorHelper` struct that performs the common tasks of synchronization of the operators, creation of channels, processing of messages and teardown. It invokes the abstractions provided by the `MessageProcessorT` trait to generate events for the lattice. 

Curent status:

- Moved `SourceOperator` and `SinkOperator` to the new abstractions.

### DIFF
--- a/src/dataflow/operator.rs
+++ b/src/dataflow/operator.rs
@@ -36,7 +36,7 @@ where
  * Sink: receives data with type T                                           *
  *****************************************************************************/
 
-pub trait Sink<S: State, T: Data>: Send {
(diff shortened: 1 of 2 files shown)


pschafhalter (approved):
LGTM

PR #154, opened 2021-03-22
Add OperatorStream struct and Stream trait

pschafhalter:


sukritkalra:
Looks good overall. Just wondering if we should rename the trait to `StreamT` and `OperatorStream` to just `Stream` since `OperatorStream` does not tell us much.

src/dataflow/connect.rs
```diff
@@ -23,14 +24,14 @@ pub fn connect_source<O, S, T>(
     operator_fn: impl Fn() -> O + Clone + Send + Sync + 'static,
     state_fn: impl Fn() -> S + Clone + Send + Sync + 'static,
     mut config: OperatorConfig,
-) -> WriteStream<T>
+) -> OperatorStream<T>
```
sukritkalra, line 27: Can we rename it to `Stream<T>`?

src/dataflow/connect.rs
```diff
@@ -71,7 +72,7 @@ pub fn connect_sink<O, S, T>(
     operator_fn: impl Fn() -> O + Clone + Send + Sync + 'static,
     state_fn: impl Fn() -> S + Clone + Send + Sync + 'static,
     mut config: OperatorConfig,
-    read_stream: &ReadStream<T>,
+    read_stream: &impl Stream<T>,
```
sukritkalra, line 75: and this to `&impl StreamT<T>`?

sukritkalra (approved):
LGTM.

PR #158, opened 2021-04-12
Fix end condition execution and maintain message statistics in streams.

sukritkalra:
- Maintains message statistics in `ReadStream` and `WriteStream`.
- Executes start conditions on the statistics of the `ReadStream` and the end conditions on the statistics of the `WriteStream`.

### DIFF
--- a/src/dataflow/deadlines.rs
+++ b/src/dataflow/deadlines.rs
@@ -12,6 +12,32 @@ use tokio::time::Duration;
 pub trait CondFn: Fn(&ConditionContext) -> bool + Send + Sync {}
 impl<F: Fn(&ConditionContext) -> bool + Send + Sync> CondFn for F {}
 
+pub struct DeadlineEvent {
+    pub stream_id: StreamId,
+    pub timestamp: Timestamp,
+    pub duration: Duration,
+    pub handler: Arc<dyn HandlerContextT>,
+    pub end_condition: Arc<dyn CondFn>,
+}
+
+impl DeadlineEvent {
+    pub fn new(
+        stream_id: StreamId,
+        timestamp: Timestamp,
+        duration: Duration,
+        handler: Arc<dyn HandlerContextT>,
+        end_condition: Arc<dyn CondFn>,
+    ) -> Self {
+        Self {
+            stream_id,
+            timestamp,
+            duration,
+            handler,
+            end_condition,
+        }
+    }
+}
+
 /// Enumerates the types of deadlines available to the user.
 pub enum Deadline {
     TimestampDeadline(TimestampDeadline),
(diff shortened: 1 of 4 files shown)


src/dataflow/deadlines.rs
```diff
@@ -12,6 +12,32 @@ use tokio::time::Duration;
 pub trait CondFn: Fn(&ConditionContext) -> bool + Send + Sync {}
 impl<F: Fn(&ConditionContext) -> bool + Send + Sync> CondFn for F {}
 
+pub struct DeadlineEvent {
```
pschafhalter, line 15: Could you add a comment on what this data structure is used for?

src/dataflow/stream/read_stream.rs
```diff
@@ -187,12 +202,58 @@ impl<D: Data> ReadStream<D> {
...
+        }
+    }
+
+    /// Clears the condition context state.
+    pub fn clear_condition_state(&mut self, stream_id: StreamId, timestamp: Timestamp) {
+        self.condition_context.clear_state(stream_id, timestamp);
+    }
+
+    /// Evaluate a condition function on the statistics of the ReadStream.
+    pub fn evaluate_condition(&self, condition_fn: &CondFn) -> bool {
```
pschafhalter, line 245: I'm not sure this is the right place to do this in terms of abstractions. Can we move this to the `OperatorExecutor` and call `get_condition_context` to evaluate the condition from there?

src/dataflow/stream/write_stream.rs
```diff
@@ -122,46 +125,68 @@ impl<D: Data> WriteStream<D> {
...
+                stats
+                    .condition_context
+                    .notify_watermark_arrival(self.id(), msg_watermark.clone());
             }
         }
         Ok(())
     }
+
+    /// Evaluates a condition function on the statistics of the write stream.
+    pub fn evaluate_condition(&self, condition_fn: &CondFn) -> bool {
```
pschafhalter, line 173: Can we also move this to the OperatorExecutor and add a `pub(crate) get_statistics(&self) -> Arc<Mutex<WriteStreamStatistics>>`?

Also, there's a chance of starvation because we're locking an `std::sync::Mutex` blocks the entire thread which can cause problems in an async context. It's fine for now and I'm not sure there's a quick fix, but worth keeping in mind.

src/node/operator_executor.rs
```diff
@@ -110,77 +145,28 @@ impl OperatorExecutorHelper {
...
-                        // duration.
-                        let queue_key: Key = self.deadline_queue.insert(
-                            (stream_id, timestamp.clone(), d.get_handler()),
-                            deadline_duration,
-                        );
-
-                        slog::debug!(
-                            crate::TERMINAL_LOGGER,
-                            "Installed a deadline handler with the Key: {:?} corresponding to \
+    fn manage_deadlines(&mut self, deadlines: Vec<DeadlineEvent>) {
```
pschafhalter, line 148: Can you add a comment on what this function does?

pschafhalter (changes requested):

src/dataflow/stream/read_stream.rs
sukritkalra: I agree, but there's a mismatch between the API on the `ReadStream` and `WriteStream` since we cannot get a `&ConditionContext` from the write stream.

I'm gonna make the change for now since these functions will remain `pub(crate)`, but once we have the rest of the deadlines implemented, we should probably revisit the correct API for this.

pschafhalter (approved):

PR #161, opened 2021-07-01
Operator Redesign [WIP]

sukritkalra:
This PR tracks the changes for the redesign of the ERDOS API that separates the `ReadOnly` and `Writeable` operators.


src/dataflow/operator.rs
```diff
@@ -43,10 +36,84 @@ where
...
 
-/*****************************************************************************
- * Sink: receives data with type T                                           *
- *****************************************************************************/
+/*************************************************************************************************
+ * ReadOnlySink: receives data with type T, and enables message callbacks to execute in parallel *
+ * by only allowing append access in the message callbacks and writable access to the state in   *
+ * the watermark callbacks.                                                                      *
+ ************************************************************************************************/
+
```
pschafhalter, line 44: Is this documentation? If so, we should comment with `///` so that the docs are properly generated.

src/dataflow/operator.rs
```diff
@@ -43,10 +36,84 @@ where
...
+    fn on_watermark(&self, ctx: &mut ReadOnlySinkContext<S, U, V>);
+}
+
+pub struct ReadOnlySinkContext<'a, S: ReadOnlyState<T, U>, T, U> {
+    pub timestamp: Timestamp,
+    pub config: OperatorConfig,
+    pub state: &'a S,
+    phantomdata_t: PhantomData<T>,
+    phantomdata_u: PhantomData<U>,
```
pschafhalter, line 61: Does the Rust compiler error if we don't include these?

src/dataflow/operator.rs
```diff
@@ -43,10 +36,84 @@ where
...
+
+    fn on_data(&self, ctx: &ReadOnlySinkContext<S, U, V>, data: &T);
+
+    fn on_watermark(&self, ctx: &mut ReadOnlySinkContext<S, U, V>);
+}
+
+pub struct ReadOnlySinkContext<'a, S: ReadOnlyState<T, U>, T, U> {
+    pub timestamp: Timestamp,
+    pub config: OperatorConfig,
+    pub state: &'a S,
```
pschafhalter, line 59: May want to hide these behind getters in a future PR.

src/dataflow/state.rs
```diff
@@ -1,25 +1,43 @@
...
 
 /// Trait that must be implemented by stream state.
 pub trait State: 'static + Clone + Send + Sync {}
 impl<T: 'static + Clone + Send + Sync> State for T {}
 
+
+// TODO (Sukrit): Do these state traits also require a way to read the state for a given timestamp?
+
+/// Trait that must be implemented by a state structure that is used in a Writeable operator.
+/// This structure must implement a `commit` method that commits the final state for a given
```
pschafhalter, line 16: Should `commit` only be accessible from within a watermark callback? Could also automatically commit after a watermark is executed.

Also, why is `commit` implemented by the user? Shouldn't the system handle time-versioned state?

src/node/lattice.rs
```diff
@@ -212,14 +214,16 @@ impl ExecutionLattice {
             let mut demoted_leaves: Vec<NodeIndex> = Vec::new();
             // These edges are redundant and must be removed.
             let mut redundant_edges: Vec<EdgeIndex> = Vec::new();
-            // Iterate through the forest with a DFS from each leaf to figure out where to add dependency edges.
+            // Iterate through the forest with a DFS from each leaf to figure out where to add
```
pschafhalter, line 217: Are these changes made by rustfmt? If so, could you upgrade Rust to >= 1.52.0 nightly to make sure our formatting is consistent?

src/node/operator_event.rs
```diff
@@ -121,45 +133,114 @@ fn resolve_access_conflicts(x: &OperatorEvent, y: &OperatorEvent) -> Ordering {
...
+        match self.operator_type {
+            OperatorType::ReadOnly => {
+                match (self.is_watermark_callback, other.is_watermark_callback) {
+                    (true, true) => {
+                        // Both of the events are watermarks, so the watermark with the lower
+                        // timestamp should run first. Ties are broken by first by dependencies to
+                        // ensure that writes to state occur before reads, and then by priority
+                        // where a smaller number is higher priority.
+                        match self.timestamp.cmp(&other.timestamp) {
+                            Ordering::Equal => match resolve_access_conflicts(&self, other) {
```
pschafhalter, line 145: Can this result in `unreachable!()`? Read-only operators should never run 2 watermarks for the same timestamp.

src/node/operator_event.rs
```diff
@@ -121,45 +133,114 @@ fn resolve_access_conflicts(x: &OperatorEvent, y: &OperatorEvent) -> Ordering {
...
+                                // Assume state is time-versioned, so dependency issues should not
+                                // arise.
+                                Ordering::Equal
+                            }
+                        }
+                    }
+                    (false, true) => other.cmp(self).reverse(),
+                    // Neither of the events are watermark callbacks.
+                    // If they have no WW, RW, or WR conflicts, they can run concurrently.
+                    (false, false) => resolve_access_conflicts(&self, other),
```
pschafhalter, line 177: Should always return `Ordering::Equal`

src/node/operator_event.rs
```diff
@@ -121,45 +133,114 @@ fn resolve_access_conflicts(x: &OperatorEvent, y: &OperatorEvent) -> Ordering {
...
+                            }
+                            Ordering::Equal => {
+                                // `self` timestamp is equal to `other`, execute `self` first.
+                                Ordering::Less
+                            }
+                            Ordering::Less => {
+                                // `self` timestamp is less than `other`, execute `self` first.
+                                Ordering::Less
+                            }
+                        }
```
pschafhalter, line 222: Since this is handled above, replace this with `other.cmp(self).reverse()`

pschafhalter (commented):
I'm not a big fan of the "writable" and "read-only" designations for operators. It might be confusing considering we already have read and write streams.

Could another name work better, e.g. Mutable and Immutable?

src/dataflow/operator.rs
sukritkalra: Nope, not documentation. Just a placeholder block to separate the different implementations. Will add documentation later once the PR is done.

src/dataflow/operator.rs
sukritkalra: Yup, "Unused parameter: consider removing `T`, referring to it in a field, or using a marker such as `PhantomData`"

src/dataflow/operator.rs
sukritkalra: Agreed. Added.

src/node/lattice.rs
sukritkalra: I'm on 1.53.0. I changed these because the original did not fit within 100 columns. I'll RustFmt this again.

src/node/operator_event.rs
sukritkalra: I think this is old code from before we had strictly defined interfaces for the operators. Our current redesign should ensure a single watermark callback for both "ReadOnly" and "Writeable" operators. Cleaning this and addressing the comments below.

src/dataflow/state.rs
sukritkalra: The API does not prevent anyone from committing inside a message callback right now since both `on_data` and `on_watermark` receive a mutable reference to `WriteableSinkContext`. We can choose to automatically commit, but I guess it would be better if the developer can commit state and continue some cleanup / message sending work.

I envision only power users implementing `commit` (or even `append` in `ReadOnlyState`). Ideally, the system would provide a default State implementation that implements `ReadOnlyState` and `WriteableState` traits, which most users would use.

The current PR does not have a system-defined implementation for these two traits.

sukritkalra:
I agree with the name changes. I am not particularly fond of `ReadOnly` and `Writeable` because they don't necessarily convey the difference between the operators. Can we come up with something that tells the user that the callbacks are serialized in one operator and that guarantee is relaxed in the other operator?

`Mutable` and `Immutable` are better than `ReadOnly` and `Writeable`, but still don't convey the difference.

src/dataflow/state.rs
pschafhalter: If that's the case, shouldn't state commits happen automatically from watermark callbacks? I'm leaning towards keeping a simpler, smaller interface since we can always expose `commit` later when a use case arises.

Also, if `commit` and `append` are intended for power-users, we should provide default implementations, e.g.
`fn commit(&self, state: &T, timestamp: &Timestamp) {};`

pschafhalter:
Let's discuss offline and try to find a better naming scheme before merging.

The changes for sinks look good.

src/node/operator_event.rs
```diff
@@ -89,77 +101,88 @@ impl PartialEq for OperatorEvent {
...
+                            }
+                            Ordering::Less => {
+                                // `self` timestamp is less than `other`, run them in any order.
+                                // Assume state is time-versioned, so dependency issues should not
+                                // arise.
+                                Ordering::Equal
+                            }
+                        }
+                    }
+                    (false, true) => other.cmp(self).reverse(),
```
ICGog, line 136: Should this compare the timestamps? other.timestamp.cmp(self.timestamp).reverse()?

src/node/operator_event.rs
```diff
@@ -89,77 +101,88 @@ impl PartialEq for OperatorEvent {
...
+                                // `self` timestamp is less than `other`, execute `self` first.
+                                Ordering::Less
+                            }
+                        }
                     }
-                    Ordering::Less => {
-                        // `self` timestamp is less than `other`, run them in any order.
-                        // Assume state is time-versioned, so dependency issues should not arise.
-                        Ordering::Equal
+                    (false, true) => other.cmp(self).reverse(),
```
ICGog, line 165: As above.

src/node/operator_executors/sink_executor.rs
```diff
@@ -0,0 +1,350 @@
...
+    fn lattice(&self) -> Arc<ExecutionLattice> {
+        Arc::clone(&self.helper.as_ref().unwrap().lattice)
+    }
+
+    fn operator_id(&self) -> OperatorId {
+        self.config.id
+    }
+}
+
+pub struct ReadOnlySinkMessageProcessor<O, S, T, U, V>
```
ICGog, line 154: I assume this is not quite ready because this processor is not used anywhere.

ICGog (commented):
Please see these comments and my previous comments.

src/dataflow/operator.rs
```diff
@@ -43,55 +36,240 @@ where
...
+    }
 }
 
-pub struct StatefulSinkContext<S: State> {
-    pub timestamp: Timestamp,
-    pub config: OperatorConfig,
-    // Hacky...
-    pub state: Arc<Mutex<S>>,
+/*************************************************************************************************
+ * WriteableSink: receives data with type T, and enables message callbacks to have mutable       *
```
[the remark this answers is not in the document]
sukritkalra: Yup, I'm gonna write docs for all the operators before I merge. This will get fixed then!

src/dataflow/state.rs
sukritkalra: Fixed in the new state implementation and the watermark callbacks now automatically commit state upon completion.

src/node/operator_executors/sink_executor.rs
sukritkalra: This was removed in the final commits.

src/node/operator_executors/mod.rs
```diff
@@ -0,0 +1,399 @@
...
+}
+
+/* ***********************************************************************************************
+ * Helper structures.
+ * ***********************************************************************************************/
+
+pub struct OperatorExecutorHelper {
+    operator_id: OperatorId,
+    lattice: Arc<ExecutionLattice>,
+    _event_runner_handles: Option<Vec<tokio::task::JoinHandle<()>>>,
```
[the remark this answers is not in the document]
sukritkalra: No idea when they got added, maybe one of us was trying something out. I removed them for now.

src/node/operator_event.rs
sukritkalra: Nope, because we don't really follow timestamp order here. Specifically, we arbitrarily break dependencies if the timestamps are equal, and choose to execute either the event in the lattice or the one being inserted.

sukritkalra:
@ICGog @pschafhalter I've addressed comments, and added some initial documentation. Can you please take a final look before we merge this in?

ICGog (commented):
What's the reason behind introducing all the different types of contexts (consistency? future feature?). It seems that three types should be enough: Context, ContextOneOut, ContextTwoOut.

sukritkalra:
I agree. I think the only reason we might want to do it is type safety so that people don't accidentally put a OneInOneOut context into a TwoInOneOut operator. I'm happy to coalesce them if there's no other major reasons.

pschafhalter (approved):
LGTM

PR #162, opened 2021-07-15
Added more mostly internal code comments.

ICGog:


src/node/lattice.rs
```diff
@@ -37,7 +37,7 @@ impl RunnableEvent {
         }
     }
 
-    /// Add a timestamp to the event.
+    /// Adds a `Timestamp` to the event.
```
pschafhalter, line 40:
```suggestion
    /// Adds a [`Timestamp`] to the event.
```

pschafhalter (commented):
See the minor change. Otherwise LGTM

src/dataflow/connect.rs
```diff
@@ -64,6 +66,8 @@ where
     write_stream
 }
 
+/// Adds a parallel sink operator, which receives data on input read streams and directly
+/// interacts with external systems.
```
sukritkalra, line 70: Can we refer to the actual types of the operators in these documentations so developers can click and see the interface?

src/dataflow/stream/errors.rs
```diff
@@ -1,6 +1,6 @@
 use crate::communication::{CommunicationError, TryRecvError};
 
-/// Errors raised by reading from a `ReadStream`.
+/// Error raised when something went wrong while reading from a `ReadStream`.
```
sukritkalra, line 3: Errors?

python/setup.py
```diff
@@ -6,7 +6,7 @@
 package_dir = os.path.dirname(__file__)
 
 # Put all package dependencies here
-requires = ["fire", "numpy"]
+requires = ["fire", "numpy==1.19.0"]
```
sukritkalra, line 9: Is there any reason we force a numpy version?

sukritkalra (commented):
I just remembered this by seeing your changes in `src/dataflow/time.rs`, but can you also please double check the `cmp` function in that file? 

Specifically, what is the correct ordering of two `Timestamp::Bottom`s or two `Timestamp::Top`s?

python/setup.py
ICGog: Yes, I'm forcing numpy to a version that still supports Python 3.6. Otherwise, our CI fails to build for Python 3.6.

ICGog: Once we fix the Rust tests, the CI should be fine again.

sukritkalra (approved):
LGTM!

PR #163, opened 2021-07-20
Timestamp deadlines [WIP]

sukritkalra:


src/dataflow/context.rs
```diff
@@ -1,11 +1,72 @@
...
+
+    /// Get the identifiers of the read streams of this operator.
+    pub fn get_read_stream_ids(&self) -> &Vec<StreamId> {
+        &self.read_stream_ids
+    }
+
+    /// Get the identifiers of the write streams of this operator.
+    pub fn get_write_stream_ids(&self) -> &Vec<StreamId> {
+        &self.write_stream_ids
+    }
```
pschafhalter, line 54: Are these used internally or do these need to be user accessible?

If they're only used internally, can we make them `pub(crate)`?

src/dataflow/context.rs
```diff
@@ -1,11 +1,72 @@
...
 };
 
+/*************************************************************************************************
+ * SetupContext: Provided to an operator's `setup` method, and allows the operators to register  *
+ * deadlines.                                                                                    *
+ ************************************************************************************************/
+
+/// A `SetupContext` is made available to an operator's `setup` method, and allows the operators to
+/// register deadlines for events along with their corresponding handlers.
+pub struct SetupContext<S> {
```
pschafhalter, line 19: Could you add a comment that `SetupContext<S>` is typed on operator state `S`?

src/dataflow/deadlines.rs
```diff
@@ -13,105 +25,78 @@ use tokio::time::Duration;
...
-        Self {
-            stream_id,
-            timestamp,
-            duration,
-            handler,
-            end_condition,
-        }
-    }
-}
+/// A trait that defines the deadline function.
```
pschafhalter, line 33: Could you describe the role of the deadline function? E.g. maps operator state and timestamp to a time point in the future?

src/dataflow/deadlines.rs
```diff
@@ -13,105 +25,78 @@ use tokio::time::Duration;
...
+pub trait DeadlineFn<S>: FnMut(&S, &Timestamp) -> Duration + Send + Sync {}
```
pschafhalter, line 34: Does it make sense to compute a duration for the deadline rather than a specific point in time at which the deadline expires (e.g. `SystemTime`)?

src/dataflow/context.rs
sukritkalra: Yup, they're internal only. Added `pub(crate)`

src/dataflow/deadlines.rs
sukritkalra: Yeah, it can go either way. It just depends on what the prevalent use-case is. I believe that it is easier to define the duration of the deadline, than an exact point in time when it expires.

@ICGog can you chime in here?

src/dataflow/deadlines.rs
```diff
@@ -1,10 +1,22 @@
...
+use crate::dataflow::{stream::StreamId, time::Timestamp, StateT};
 use std::{
     collections::{HashMap, HashSet},
     sync::{Arc, Mutex},
 };
 use tokio::time::Duration;
 
+/*************************************************************************************************
+ * DeadlineId: The identifier assigned to each deadline in order to enable handler invocation.   *
+ ************************************************************************************************/
```
pschafhalter, line 10: Not sure we need this since it duplicates the documentation below.

src/dataflow/deadlines.rs
```diff
@@ -1,10 +1,22 @@
...
+
+/// The identifier assigned to each deadline installed in a SetupContext.
+pub type DeadlineId = crate::Uuid;
+
+/*************************************************************************************************
+ * CondFn: The type of the start and end condition functions invoked by the different types of   *
+ * deadlines to decide when to initiate and finish the deadline.                                 *
+ ************************************************************************************************/
```
pschafhalter, line 18: Same here, not sure this is necessary since we already have documentation above the `CondFn` trait.

src/dataflow/deadlines.rs
```diff
@@ -13,105 +25,80 @@ use tokio::time::Duration;
...
-    pub stream_id: StreamId,
-    pub timestamp: Timestamp,
-    pub duration: Duration,
-    pub handler: Arc<Mutex<dyn HandlerContextT>>,
-    pub end_condition: Arc<dyn CondFn>,
-}
+/*************************************************************************************************
+ * DeadlineFn: A DeadlineFn defines the duration between the start and the end condition that an *
+ * operator should be restricted to.                                                             *
+ ************************************************************************************************/
```
pschafhalter, line 31: Also don't think it's necessary to duplicate the documentation.

src/dataflow/deadlines.rs
```diff
@@ -13,105 +25,80 @@ use tokio::time::Duration;
...
-impl ConditionContext {
-    pub fn new() -> Self {
-        ConditionContext {
-            message_count: HashMap::new(),
-            watermark_status: HashMap::new(),
-        }
-    }
+/*************************************************************************************************
+ * Deadline: Define the different types of deadlines available to operators.                     *
+ ************************************************************************************************/
```
pschafhalter, line 49: Can we convert this to documentation (comments with 3 slashes) and move it to just above the Deadline trait?

src/dataflow/deadlines.rs
```diff
@@ -13,105 +25,80 @@ use tokio::time::Duration;
...
+
+    fn invoke_handler(&self, state: &mut S, timestamp: &Timestamp);
 }
 
-pub struct TimestampDeadline {
+/*************************************************************************************************
+ * TimestampDeadline: A deadline that constrains the duration between the start and the end      *
+ * conditons for a particular timestamp.                                                         *
+ ************************************************************************************************/
+
```
pschafhalter, line 73: Convert to documentation (3 slashes).

src/main.rs
```diff
@@ -59,7 +69,42 @@ impl StateT for SquareOperatorState {
...
+        Duration::new(5, 0)
+    }
+}
+
 impl OneInOneOut<SquareOperatorState, usize, usize> for SquareOperator {
+    fn setup(&mut self, ctx: &mut SetupContext<SquareOperatorState>) {
+        let deadline_ctx = Arc::clone(&self.deadline_ctx);
+        ctx.add_deadline(TimestampDeadline::new(
+            move |_s: &SquareOperatorState, _t: &Timestamp| -> Duration {
+                deadline_ctx.lock().unwrap().deadline()
```
pschafhalter, line 96: I'm not a fan of the user manually locking state. Is there anyway we can "hide" the context or pass it to the system via the SetupContext?

sukritkalra: The user doesn't manually lock the operator state, but this example shows that users can install deadline definition functions and still have state for them separately, if need be. 

The other way of doing it is to force everyone to define a structure that implements a `DeadlineContext` trait (just like we used to have before).

src/dataflow/deadlines.rs
ICGog: Agreed, this can go either way. I'm tempted to treat these as relative deadlines for now (i.e., compute duration) rather than absolute deadlines because this code implements timestamp deadlines.

src/dataflow/deadlines.rs
```diff
@@ -189,6 +135,109 @@ impl TimestampDeadline {
             current_timestamp,
             condition_context,
         );
+        // TODO (Sukrit): Implement the end condition function.
```
ICGog, line 138: Don't we need this to be implemented in this PR for the TimestampDeadline to work?

src/node/operator_executors/mod.rs
```diff
@@ -91,32 +98,12 @@ where
     /// Generates a DeadlineEvent for arming a deadline.
     fn arm_deadlines(
         &self,
-        setup_context: &dyn SetupContextT,
-        read_stream: &ReadStream<T>,
-        timestamp: Timestamp,
+        _setup_context: &mut SetupContext<S>,
```
ICGog, line 101: Why do you still need this method?

src/main.rs
ICGog: Yes, I'm also not a fan of locking the deadline contexts. What is the argument against implementing the DeadlineContext trait?

ICGog:
Ping

pschafhalter (approved):
LGTM.

I'm personally not a huge fan of the state API -- especially that the user needs to implement `get_last_committed_timestamp`, but IMO we can merge this and clean it up later.

pschafhalter:
Also could you fix the merge conflicts?

PR #165, opened 2021-08-18
Run IngestStreams and ExtractStreams on Node 0

pschafhalter:
Modifies IngestStreams and ExtractStreams to run on node 0, which is a first step towards providing a single driver for ERDOS programs.

As a temporary solution, the dataflow graph places ingest and extract streams on node 0. I'll improve this abstraction in a follow-up PR in which I'm planning to refactor the dataflow graph.


src/dataflow/stream/ingest_stream.rs
```diff
@@ -68,43 +65,33 @@ where
     // TODO (Sukrit) :: Why does this constructor require the NodeID? Inconsistency between
     // constructors of different types of streams.
```
sukritkalra, line 66: Can you remove this TODO now?

sukritkalra (approved):
LGTM. Minor comment.

PR #166, opened 2021-08-23
Redesigned Python API

sukritkalra:


python/erdos/streams.py
```diff
@@ -260,24 +213,23 @@ class ExtractStream(object):
     ERDOS applications.
 
     The driver can initialize a new :py:class:`ExtractStream` by passing the
-    instance of :py:class:`ReadStream` returned by :py:func:`connect`. Similar
+    instance of :py:class:`tream` returned by :py:func:`connect`. Similar
```
ICGog, line 216: Typo

ICGog (commented):
Comments from a partial review.

python/erdos/timestamp.py
```diff
@@ -43,10 +87,19 @@ def __ne__(self, timestamp):
         return not self.__eq__(timestamp)
 
     def __lt__(self, timestamp):
+        # Compare Bottom timestamps.
+        if self.is_bottom:
```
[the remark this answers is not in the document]
sukritkalra: Fixed.

python/erdos/operator.py
```diff
@@ -6,85 +6,10 @@
...
-
-    Instead of ERDOS invoking callbacks registered on the read streams, an
-    operator can also take control of the execution by overriding the
-    :py:func:`Operator.run` method as follows::
-
-        class MapOperator(erdos.Operator):
-            def run(self):
-                message = self.read_stream.read()
-                # Work on the message.
+class BaseOperator(object):
```
pschafhalter, line 9: Can the classes in this module inherit from [abstract base class](https://docs.python.org/3/library/abc.html)?

python/erdos/operator.py
```diff
@@ -127,6 +52,333 @@ def save_trace_events(self, file_name):
...
+    def run(self, write_stream):
+        """Runs the operator.
+
+        Invoked automatically by ERDOS, and provided with a
+        :py:class:`WriteStream` to send data on.
+
+        Args:
+            write_stream: A :py:class:`WriteStream` instance to send data on.
+        """
+        pass
```
pschafhalter, line 83: API question: should we force the user to implement all of these methods when writing new operators (e.g. by raising `NotImplementedError` or marking them as abstract methods?

An advantage is that the system will catch bugs, e.g. from mispelling the method name when overriding. A disadvantage is that a new operator must implement all of these methods.

python/erdos/timestamp.py
```diff
@@ -1,37 +1,81 @@
...
+                self._is_top = is_top
+                self._is_bottom = is_bottom
+                self.coordinates = coordinates
+            elif coordinates is not None and not is_bottom and not is_top:
+                self._is_top = is_top
+                self._is_bottom = is_bottom
+                self.coordinates = coordinates
+            else:
+                raise ValueError("Timestamp should either have coordinates"
+                                 "or be either Top or Bottom")
```
pschafhalter, line 53: Might be easier to maintain if we always instantiate a `PyTimestamp`, and store `self._py_timestamp`.

Then we can re-use the Rust implementation for `__lt__`, `__leq__` by writing it in Rust in the `PyTimestamp` class. Properties like `is_top` would return `self._py_timestamp.is_top`, etc.

pschafhalter: We might also want to consider moving the `Timestamp` class to pure Rust and getting rid of this Python wrapper. I believe pyo3 allows us to add docstrings from Rust now too.

Thoughts?

src/python/py_timestamp.rs
```diff
@@ -0,0 +1,78 @@
...
+            Timestamp::Top => Self {
+                coordinates: None,
+                is_top: true,
+                is_bottom: false,
+            },
+        }
+    }
+}
+
+impl Into<Timestamp> for PyTimestamp {
```
pschafhalter, line 68: It's [better practice](https://doc.rust-lang.org/std/convert/trait.From.html) to `impl From<PyTimestamp> for Timestamp`

pschafhalter (commented):
Why do we need separate Python executors? Can't we implement `PySource`, `PySink`, `PyOneInOnOut`, etc. operators? These would take a reference to the Python operator as an argument, and be responsible for calling the correct methods on the Python operator object.

In addition, this would also let us use the Rust `connect_*` functions, rather than duplicating much of the logic in `python/mod.rs`.

I'm in general pretty concerned about the amount of logic duplication going on between the Python and Rust APIs. A lot of duplication makes it difficult to maintain the codebase and introduce new features.

src/python/py_timestamp.rs
sukritkalra: Thanks! I'll change this to a `From` in this file instead of having it in `src/dataflow/time.rs` though.

python/erdos/__init__.py
```diff
@@ -304,17 +415,15 @@ def wrapper(*args, **kwargs):
...
     "ExtractStream",
-    "Operator",
-    "OperatorConfig",
     "Profile",
     "Message",
     "WatermarkMessage",
     "Timestamp",
-    "connect",
+    "connect_source",
+    "connect_sink",
```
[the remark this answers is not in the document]
sukritkalra: Yes, added them in the latest changes.

python/erdos/timestamp.py
sukritkalra: I wrote a minor PyTimestamp wrapper that patches almost all methods from Python to Rust so we utilize our implementations in Rust now for comparisons.

python/erdos/operator.py
sukritkalra: I added a logger warning if the user leaves all of the methods of an operator empty.

Cargo.toml
```diff
@@ -27,7 +27,7 @@ futures-intrusive = "0.4"  # Required for futures-delay-queue
 futures-util = "0.3.5"
 lazy_static = "1.4.0"
 petgraph = "0.5.0"
-pyo3 = { version = "0.8.2", features = ["unsound-subclass"], optional = true }
+pyo3 = { version = "0.14.1", optional = true }
```
sukritkalra, line 30: I need to bump this before merging.

pschafhalter (approved):
LGTM, once the merge conflict is fixed and PyO3 is bumped to 0.14.3.

PR #175, opened 2021-10-05
Update watermark example, Fix Timestamp bug

objorkman:
Updates to `watermarks.py` for redesign, and fix in `streams.py` when parsing a message that is a watermark.

### DIFF
--- a/python/erdos/streams.py
+++ b/python/erdos/streams.py
@@ -20,7 +20,7 @@ def _parse_message(internal_msg):
         return pickle.loads(internal_msg.data)
     if internal_msg.is_watermark():
         return WatermarkMessage(
-            Timestamp(coordinates=internal_msg.timestamp,
+            Timestamp(_py_timestamp=internal_msg.timestamp,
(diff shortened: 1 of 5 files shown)


python/erdos/streams.py
```diff
@@ -20,7 +20,7 @@ def _parse_message(internal_msg):
         return pickle.loads(internal_msg.data)
     if internal_msg.is_watermark():
         return WatermarkMessage(
-            Timestamp(coordinates=internal_msg.timestamp,
+            Timestamp(_py_timestamp=internal_msg.timestamp,
                       is_top=internal_msg.is_top_watermark()))
```
pschafhalter, line 24: Delete this line. `PyTimestamp` specifies whether a timestamp is top.

pschafhalter (commented):
The example look good, and nice catch on the bug!

Since this is an example, could you add type annotations for the arguments in the operator methods? That way, the examples will serve as a better reference for how to write ERDOS applications.

pschafhalter:
LGTM! Nice work!

PR #181, opened 2021-11-01
[Python] Add ID property to ReadStream, WriteStream

objorkman:
* Examples are tested to be working
* Add read-only ID property to py_read_stream and py_write_stream


pschafhalter (approved):
LGTM

PR #182, opened 2021-11-01
[Python] Add ID property to ExtractStream, IngestStream

objorkman:
* Similar changes like #181 
* Examples tested to work


python/erdos/streams.py
```diff
@@ -15,7 +15,7 @@ def _parse_message(internal_msg):
     """Creates a Message from an internal stream's response.
 
     Args:
-        internal_msg (PyMessage): The internal message to parse.
+        internal_msg (PyMessage): The int   ernal message to parse.
```
pschafhalter, line 18:
```suggestion
        internal_msg (PyMessage): The internal message to parse.
```

pschafhalter (approved):
LGTM

PR #184, opened 2021-11-10
[Bug] Fix possible deadlock in operator executors

pschafhalter:
Deadlocks could occur because functions generating operator events for messages and callbacks sometimes acquire locks in different orders.

## Example

The message callback in `OneInOneOutExceutor` acquires a lock on the operator before the state:
https://github.com/erdos-project/erdos/blob/54c2a4db7bbef272de8ad09e578a8188ddae0f3d/src/node/operator_executors/one_in_one_out_executor.rs#L344-L354

Whereas the watermark callback acquires a lock on the state before the operator:
https://github.com/erdos-project/erdos/blob/54c2a4db7bbef272de8ad09e578a8188ddae0f3d/src/node/operator_executors/one_in_one_out_executor.rs#L406-L421

### DIFF
--- a/src/node/operator_executors/one_in_one_out_executor.rs
+++ b/src/node/operator_executors/one_in_one_out_executor.rs
@@ -342,13 +342,12 @@ where
             HashSet::new(),
             HashSet::new(),
             move || {
-                operator.lock().unwrap().on_data(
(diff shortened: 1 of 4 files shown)


sukritkalra (approved):
LGTM! Good catch. 👍

PR #185, opened 2021-11-10
Switch logging to tracing from slog

pschafhalter:
Switches logging to [tracing](https://github.com/tokio-rs/tracing) from [slog](https://github.com/slog-rs/slog).

## Additional Changes
- More customization for logging via the Node configuration and command line interfaces.
- Rename `num_worker_threads` to `num_threads` to distinguish from the worker tasks.

## Advantages of tracing
- Supports [spans](https://docs.rs/tracing/0.1.29/tracing/#spans) for monitoring the progress/duration code contexts. I've found this useful for concurrent debugging.
- Tracing integrates with Tokio and provides support for async contexts.
- Don't need to pass a logger when recording events.
- Support for additional logging interfaces (e.g. writing to file, integrating with [OpenTelemetry](https://github.com/tokio-rs/tracing/tree/master/tracing-opentelemetry)).
- Easy to add additional information such as current OS thread.

## Advantages of slog
- Slightly better measured performance
   - 1.3 us/event for slog with slog-async-drain vs. 3 us/event for tracing with tracing-appender.
   - In these measurements, the slog's asynchronous logger dropped most events.
- Users can instantiate that are independent of ERDOS.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,13 +25,13 @@ futures = "0.3.5"
 futures-delay-queue = "0.4.2"
 futures-intrusive = "0.4"  # Required for futures-delay-queue
 futures-util = "0.3.5"
-lazy_static = "1.4.0"
 petgraph = "0.5.0"
 pyo3 = { version = "0.14.3", optional = true }
 rand = "0.3"
 serde = { version = "1.0.115", features = ["derive"] }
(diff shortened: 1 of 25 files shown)


ICGog (approved):
LGTM

PR #188, opened 2021-12-19
[Rust] Add ROS Operators

objorkman:
* Adds `to_ros_operator` and `from_ros_operator` which operate on generic types
* Examples for how to use each placed in new `examples/` directory

Known issues:
* Each operator has a `rosrust::init()`, so only 1 operator can be used in an erdos application. Workaround is to only initialize once globally, library allows multiple nodes running on the same process.
* `from_erdos` example publishes rosrust::RawMessage, which is a serialized version of the original String


pschafhalter (changes requested):
Great work on the implementation! Needs some polishing before we can merge:
1. Can you rebase on the latest commits to redesign? There's changes to logging (moved from slog to tracing).
2. Please add more documentation to the operators and the examples! Imagine that this will be used by someone who doesn't know much about ERDOS and ROS, but wants to quickly connect them without running into bugs. The documentation should provide clear information on how to achieve this goal, and explain the necessary steps (e.g. how do I write a `to_ros_msg` function). More comments in the examples would help too -- imagine that you're walking me through the code.
3. Can multiple to ROS and from ROS operators can run simultaneously in the same process?
4. Can we hide the ROS integration behind a [feature flag](https://doc.rust-lang.org/cargo/reference/features.html)? This'll improve compile times when we're not editing the ROS integration (e.g. we already hide Python integration behind a feature). The best way to go about this is to move the ROS operator files into a `ros` submodule. Create a folder called `ros` under the `operators` folder and add `mod.rs` that imports the operators like in the current `operators/mod.rs`. You'll also need to edit `Cargo.toml`. In `operators/mod.rs`, add a section for public submodules and to `operators/mod.rs` add the following:
```rust
// Public submodules
#[cfg(feature = "ros")]    
pub mod ros;
```

examples/to_erdos.rs
```diff
@@ -0,0 +1,83 @@
...
+use erdos::dataflow::context::*;
+use erdos::dataflow::operator::*;
+use erdos::dataflow::operators::*;
+use erdos::dataflow::*;
+use erdos::dataflow::Message;
+use erdos::node::Node;
+use erdos::Configuration;
+
+/// Subscribes to ROS topic named "chatter" which consists of String messages. 
+/// Messages are converted and sent on an erdos stream which is captured by a Sink node.
```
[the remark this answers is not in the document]
objorkman: I placed the comment there to follow in the likes of `join_operator` and `map_operator`. Can move it though!

src/dataflow/operators/from_ros_operator.rs
```diff
@@ -0,0 +1,56 @@
...
+        let to_erdos_msg = self.to_erdos_msg.clone();
+        let w_s = Arc::new(Mutex::new(write_stream.clone()));
+
+        let _subscriber_raii = rosrust::subscribe(self.topic.as_str(), 2, move |v: T| {
+            let converted = (to_erdos_msg)(&v);
+            slog::info!(
+                crate::TERMINAL_LOGGER,
+                "FromRosOperator: Received and Converted {:?}",
+                converted,
+            );
```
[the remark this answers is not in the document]
objorkman: Updated `ToRosOperator` with this. Is there a way to access the operator name without the context, which is the case in the run function of `FromRosOperator`?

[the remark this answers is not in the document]
pschafhalter: Not right now. We probably need to create a context for the run function as well (cc @sukritkalra)

examples/from_erdos.rs
```diff
@@ -0,0 +1,69 @@
...
+
+
+fn main() {
+    let args = erdos::new_app("ERDOS").get_matches();
+    let mut node = Node::new(Configuration::from_args(&args));
+
+    let source_config = OperatorConfig::new().name("SourceOperator");
+    let source_stream = erdos::connect_source(SourceOperator::new, || {}, source_config);
+
+    let erdos_to_ros = |input: &String| -> rosrust::RawMessage {
```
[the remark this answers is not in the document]
objorkman: It doesn't look like `rosrust::MsgValue` implements the `rosrust::Message` trait, but all of the messages in `rosrust_msg` do. I can use `rosrust_msg::std_msgs::String` instead like is used in the `ros_to_erdos` conversion.

objorkman:
Thanks for the comments Peter! I made changes that hopefully addressed everything and added documentation+examples. 

Regarding running multiple ROS operators in the same process, we briefly discussed this but was setting a global flag the way to go, and only the first created operator calling `rosrust::init` the best way to go?

src/dataflow/operators/ros/from_ros_operator.rs
```diff
@@ -0,0 +1,70 @@
...
+/// for a variety of supported standard ROS messages.
+/// 
+/// # Example
+/// The following example shows a conversion function which takes a [`rosrust_msg::sensor_msgs::Image`]
+/// and returns an ERDOS message containing [`Vec<u8>`] a vector of bytes.
+/// 
+/// ```
+/// fn ros_to_erdos(input: &rosrust_msg::sensor_msgs::Image) -> Message<Vec<u8>> {
+///     Message::new_message(Timestamp::Time(vec![0 as u64]), input.data)
+/// }
```
pschafhalter, line 22: Could you also include an example that initializes a `FromRosOperator` with the function?

src/dataflow/operators/ros/from_ros_operator.rs
```diff
@@ -0,0 +1,70 @@
...
+/// Conversion function should convert a ROS type which implements the [`rosrust::Message`] 
+/// trait and return a ERDOS [`Message`] containing data of a Rust data type. See [`rosrust_msg`]
+/// for a variety of supported standard ROS messages.
+/// 
+/// # Example
+/// The following example shows a conversion function which takes a [`rosrust_msg::sensor_msgs::Image`]
+/// and returns an ERDOS message containing [`Vec<u8>`] a vector of bytes.
+/// 
+/// ```
+/// fn ros_to_erdos(input: &rosrust_msg::sensor_msgs::Image) -> Message<Vec<u8>> {
```
pschafhalter, line 20:
```suggestion
/// fn ros_image_to_bytes(input: &rosrust_msg::sensor_msgs::Image) -> Message<Vec<u8>> {
```

src/dataflow/operators/ros/to_ros_operator.rs
```diff
@@ -0,0 +1,67 @@
...
+/// 
+/// Conversion function should convert a Rust data type and return a ROS type which implements 
+/// the [`rosrust::Message`] trait.
+/// 
+/// # Example
+/// The following example shows a conversion function which takes a Rust [`i32`] and converts it
+/// to a ROS message with [`rosrust_msg::std_msgs::Int32`] data. 
+/// 
+/// ```
+/// fn erdos_to_ros(input: &i32) -> rosrust_msg::std_msgs::Int32 {
```
pschafhalter, line 19:
```suggestion
/// fn erdos_int_to_ros_int(input: &i32) -> rosrust_msg::std_msgs::Int32 {
```

examples/to_erdos.rs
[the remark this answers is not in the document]
pschafhalter: Please move it to the top!

src/dataflow/operators/ros/to_ros_operator.rs
```diff
@@ -16,7 +16,7 @@ use std::sync::Arc;
 /// to a ROS message with [`rosrust_msg::std_msgs::Int32`] data. 
 /// 
 /// ```
-/// fn erdos_to_ros(input: &i32) -> rosrust_msg::std_msgs::Int32 {
+/// fn erdos_int_to_ros_int(input: &i32) -> rosrust_msg::std_msgs::Int32 {
 ///     rosrust_msg::std_msgs::Int32 { data: input.data }
 /// }
```
pschafhalter, line 21: Could you add an example that initializes a `ToRosOperator` as well?

pschafhalter (approved):
LGTM

src/dataflow/operators/ros/to_ros_operator.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl<T, U: rosrust::Message> ToRosOperator<T, U>
+where
+    T: Data + for<'a> Deserialize<'a>,
+{
+    pub fn new<F>(topic: &str, to_ros_msg: F) -> Self
+    where
+        F: 'static + Fn(&T) -> U + Send + Sync,
+    {
+        rosrust::init("publisher");
```
sukritkalra, line 49:
```suggestion
        rosrust::init("ERDOS Publisher");
```

So we can easily figure out the nodes that belongs to ERDOS. 

Also, since there may be multiple of these nodes that come out of a single ERDOS dataflow graph, we might run into naming conflicts for ROS nodes here. Can you check what the ROS documentation says about node name conflicts?

src/dataflow/operators/ros/to_ros_operator.rs
```diff
@@ -0,0 +1,76 @@
...
+        let msg = (self.to_ros_msg)(data);
+
+        tracing::info!(
+            "{} @ {:?}: Sending {:?}",
+            ctx.get_operator_config().get_name(),
+            timestamp,
+            msg,
+        );
+        // Publishes converted message on topic.
+        self.publisher.send(msg).unwrap();
```
sukritkalra, line 72: This might be a topic for future discussion, but I'm wondering if we should check if there are any subscribers on the topic that we're publishing? ERDOS aims to ensure a dataflow execution that is free of any message loss. However, this guarantee breaks down once we touch the ROS interface.

We may also be able to require the number of subscribers that an ERDOS operator expects to have to ensure that there is no message loss.

Caveat: Even with these changes, we still cannot guarantee a loss-free communication since ROS can choose to arbitrarily drop messages from its queue at the Subscriber end.

src/dataflow/operators/ros/to_ros_operator.rs
```diff
@@ -0,0 +1,76 @@
...
+where
+    T: Data + for<'a> Deserialize<'a>,
+{
+    pub fn new<F>(topic: &str, to_ros_msg: F) -> Self
+    where
+        F: 'static + Fn(&T) -> U + Send + Sync,
+    {
+        rosrust::init("publisher");
+        Self {
+            publisher: rosrust::publish(topic, 2).unwrap(),
```
sukritkalra, line 51: Why was the topic length set to 2?

src/dataflow/operators/ros/to_ros_operator.rs
```diff
@@ -0,0 +1,76 @@
...
+///     &source_stream,
+/// );
+/// ```
+
+pub struct ToRosOperator<T, U: rosrust::Message>
+where
+    T: Data + for<'a> Deserialize<'a>,
+{
+    publisher: rosrust::Publisher<U>,
+    to_ros_msg: Arc<dyn Fn(&T) -> U + Send + Sync>,
```
sukritkalra, line 38: I think we should expose the timestamp to the `to_ros_msg` function. Many ROS types have a separate timestamp field, and we want to encourage the flow of timestamps to ROS too.

Developers can just disregard it if they don't need it. But right now, they're gonna have to duplicate the timestamp in the data to get access to it in the `to_ros_msg` function.

src/dataflow/operators/ros/to_ros_operator.rs
```diff
@@ -0,0 +1,76 @@
...
+            "{} @ {:?}: Sending {:?}",
+            ctx.get_operator_config().get_name(),
+            timestamp,
+            msg,
+        );
+        // Publishes converted message on topic.
+        self.publisher.send(msg).unwrap();
+    }
+
+    fn on_watermark(&mut self, _ctx: &mut SinkContext<()>) {}
```
sukritkalra, line 75: Similar in spirit to the exposing of timestamps to the `to_ros_msg` function, we should be able to send watermarks downstream. 

Although this might be a more difficult endeavor, does ROS have Enum typed messages?

src/dataflow/operators/ros/from_ros_operator.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl<T: rosrust::Message, U> FromRosOperator<T, U>
+where
+    U: Data + for<'a> Deserialize<'a>,
+{
+    pub fn new<F>(topic: &str, to_erdos_msg: F) -> Self
+    where
+        F: 'static + Fn(&T) -> Message<U> + Send + Sync,
+    {
+        rosrust::init("subscriber");
```
sukritkalra, line 53:
```suggestion
        rosrust::init("ERDOS Subscriber");
```

See comment below for node name clashes in ROS.

src/dataflow/operators/ros/from_ros_operator.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl<T: rosrust::Message, U> Source<(), U> for FromRosOperator<T, U>
+where
+    U: Data + for<'a> Deserialize<'a>,
+{
+    fn run(&mut self, write_stream: &mut WriteStream<U>) {
+        let to_erdos_msg = self.to_erdos_msg.clone();
+        let w_s = Arc::new(Mutex::new(write_stream.clone()));
+
+        let _subscriber_raii = rosrust::subscribe(self.topic.as_str(), 2, move |v: T| {
```
sukritkalra, line 69: Same comment about topic length as before.

Why was the variable named `v`? Can we change this to something more descriptive like `ros_msg`?

src/dataflow/operators/ros/from_ros_operator.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+}
+
+impl<T: rosrust::Message, U> Source<(), U> for FromRosOperator<T, U>
+where
+    U: Data + for<'a> Deserialize<'a>,
+{
+    fn run(&mut self, write_stream: &mut WriteStream<U>) {
+        let to_erdos_msg = self.to_erdos_msg.clone();
+        let w_s = Arc::new(Mutex::new(write_stream.clone()));
```
sukritkalra, line 67: Can you change the variable name from `w_s` to something more descriptive? Maybe `write_stream_clone`?

src/dataflow/operators/ros/to_ros_operator.rs
```diff
@@ -0,0 +1,76 @@
...
+    fn on_data(&mut self, ctx: &mut SinkContext<()>, data: &T) {
+        let timestamp = ctx.get_timestamp().clone();
+        let msg = (self.to_ros_msg)(data);
+
+        tracing::info!(
+            "{} @ {:?}: Sending {:?}",
+            ctx.get_operator_config().get_name(),
+            timestamp,
+            msg,
+        );
```
sukritkalra, line 70: This is `info` and the logs in `FromRosOperator` are `debug`. 

My opinion is to change them both to `trace`. @pschafhalter can you chime in here?

src/dataflow/operators/ros/to_ros_operator.rs
```diff
@@ -0,0 +1,76 @@
...
+
+pub struct ToRosOperator<T, U: rosrust::Message>
+where
+    T: Data + for<'a> Deserialize<'a>,
+{
+    publisher: rosrust::Publisher<U>,
+    to_ros_msg: Arc<dyn Fn(&T) -> U + Send + Sync>,
+}
+
+impl<T, U: rosrust::Message> ToRosOperator<T, U>
```
sukritkalra, line 41: Nit: `ToROSOperator`?

src/dataflow/operators/ros/from_ros_operator.rs
```diff
@@ -0,0 +1,79 @@
...
+
+        let _subscriber_raii = rosrust::subscribe(self.topic.as_str(), 2, move |v: T| {
+            let converted = (to_erdos_msg)(&v);
+            tracing::debug!("FromRosOperator: Received and Converted {:?}", converted,);
```
sukritkalra, line 71: See comment below about log level mismatches.

src/dataflow/operators/ros/from_ros_operator.rs
```diff
@@ -0,0 +1,79 @@
...
+impl<T: rosrust::Message, U> Source<(), U> for FromRosOperator<T, U>
+where
+    U: Data + for<'a> Deserialize<'a>,
+{
+    fn run(&mut self, write_stream: &mut WriteStream<U>) {
+        let to_erdos_msg = self.to_erdos_msg.clone();
+        let w_s = Arc::new(Mutex::new(write_stream.clone()));
+
+        let _subscriber_raii = rosrust::subscribe(self.topic.as_str(), 2, move |v: T| {
+            let converted = (to_erdos_msg)(&v);
```
sukritkalra, line 70:
```suggestion
            let erdos_msg = (to_erdos_msg)(&ros_msg);
```

src/dataflow/operators/ros/from_ros_operator.rs
```diff
@@ -0,0 +1,79 @@
...
+///     move || -> FromRosOperator<rosrust_msg::sensor_msgs::Image, Vec<u8>> {
+///         FromRosOperator::new("image_topic", ros_image_to_bytes)
+///     },
+///     || {},
+///     ros_source_config,
+/// );
+/// ```
+
+#[derive(Clone)]
+pub struct FromRosOperator<T: rosrust::Message, U>
```
sukritkalra, line 37: Nit: `FromROSOperator`

src/dataflow/operators/ros/from_ros_operator.rs
```diff
@@ -0,0 +1,79 @@
...
+{
+    topic: String,
+    to_erdos_msg: Arc<dyn Fn(&T) -> Message<U> + Send + Sync>,
+}
+
+impl<T: rosrust::Message, U> FromRosOperator<T, U>
+where
+    U: Data + for<'a> Deserialize<'a>,
+{
+    pub fn new<F>(topic: &str, to_erdos_msg: F) -> Self
```
sukritkalra, line 49: One potential way of solving the unavailability of names in the `run` method is to require the developers to provide the `OperatorConfig` in the `new` method. You can also probably use that to sidestep the ROS node name clash issues discussed below.

sukritkalra (changes requested):
Looks great. Thanks for the PR! My comments are more towards the general API and guarantees that we may think of providing at some point.

src/dataflow/operators/ros/to_ros_operator.rs
pschafhalter: Let's set it to 1024? Could you make this a constant in `operators/ros/mod.rs`?

```rust
const ROS_QUEUE_SIZE = 1024;
```

src/dataflow/operators/ros/to_ros_operator.rs
pschafhalter: Agree with this. Can we change the `to_ros_msg` type to take ERDOS messages, e.g. `Fn(&Message<T>) -> U`?

src/dataflow/operators/ros/to_ros_operator.rs
pschafhalter: Agreed, let's move both to trace.

src/dataflow/operators/ros/to_ros_operator.rs
pschafhalter: Let's go with `ToRosOperator`. Looks like it fits better with typical camel case usage:
https://stackoverflow.com/questions/15526107/acronyms-in-camelcase

src/dataflow/operators/ros/from_ros_operator.rs
pschafhalter: Same as above, let's stick with `FromRosOperator`

src/dataflow/operators/ros/to_ros_operator.rs
objorkman: The rosrust library we're using allows for this -- See this [example](https://github.com/adnanademovic/rosrust/blob/3c5edfaa0d5c2f7528a50966f2ff6c1bb1bf7ea1/rosrust/examples/publisher.rs#L35).

src/dataflow/operators/ros/to_ros_operator.rs
objorkman: It seems like other client libraries `rospy` and `roscpp` will shutdown the older node with the name conflict. Upon testing, this library panics and crashes.

src/dataflow/operators/ros/to_ros_operator.rs
objorkman: ROS/ROS2 don't support Enum typed messages. [Here's](https://answers.ros.org/question/9427/enum-in-msg/) a workaround using constants.

examples/from_erdos.rs
```diff
@@ -0,0 +1,76 @@
+/// Publishes erdos messages of type String to ROS topic "chatter".
```
pschafhalter, line 1: Can we rename this file `erdos_to_ros.rs`?

examples/to_erdos.rs
```diff
@@ -0,0 +1,60 @@
+/// Subscribes to ROS topic named "chatter" which consists of String messages.
```
pschafhalter, line 1: Can we rename this file `ros_to_erdos.rs`?

src/dataflow/operators/ros/from_ros_operator.rs
```diff
@@ -0,0 +1,80 @@
...
+/// );
+/// ```
+
+#[derive(Clone)]
+pub struct FromRosOperator<T: rosrust::Message, U>
+where
+    U: Data + for<'a> Deserialize<'a>,
+{
+    topic: String,
+    to_erdos_msg: Arc<dyn Fn(&T) -> Message<U> + Send + Sync>,
```
pschafhalter, line 43: Can we rename this to `from_ros_msg`?

src/dataflow/operators/ros/to_ros_operator.rs
pschafhalter: Maybe we can have the`to_ros_msg` function handle watermarks as well, since it already takes an ERDOS message? One potential issue is that each watermark must produce a ROS message, since the function type is `&erdos_msg -> ros_msg`.

I propose we change the function type to `&erdos_message -> Vec<ros_message>`. This way, we have maximum flexibility because both watermarks and data messages could produce 0, 1, or several ERDOS messages.

Likewise, to match the API, I also propose we change the function type for `to_erdos_msg` (which should be renamed to `from_ros_msg`) to `&ros_msg -> Vec<erdos_msg>`. This will support useful patterns like sending data followed by a watermark.

@sukritkalra @objorkman thoughts?

sukritkalra: As discussed in the ERDOS meeting, we are in agreement on this change. Let's make sure that both `to_ros_msg` and `from_ros_msg` take `Vec<Message<T>>` and that the same method is invoked for both `Message` and `WatermarkMessage` callbacks here.

sukritkalra (approved):
LGTM! Thanks for the PR!

PR #192, opened 2022-01-26
Speed up CI

pschafhalter:
Fixes the CI rust toolchain version and uses [rust-cache](https://github.com/Swatinem/rust-cache) to cache dependencies and speed up CI runs.


sukritkalra (approved):
LGTM!

PR #193, opened 2022-01-26
[LINQ] Add example, and traits for filter, map, split

sguduguntla:
Here are just the Rust LINQ Operators

PR #196, opened 2022-01-26
[LINQ] Add concat operator

pschafhalter:


erdos/src/dataflow/operators/concat.rs
```diff
@@ -0,0 +1,91 @@
...
+
+impl<D> Concat<D> for Stream<D>
+where
+    D: Data + for<'a> Deserialize<'a>,
+{
+    fn concat(&self, other: &Stream<D>) -> Stream<D> {
+        crate::connect_two_in_one_out(
+            ConcatOperator::new,
+            || {},
+            OperatorConfig::new().name("ConcatOperator"),
```
sukritkalra, line 86: Can we add the name of the two streams to this operator for future debugging purposes?

sukritkalra (approved):
LGTM. Minor comments!

erdos/src/dataflow/operators/concat.rs
```diff
@@ -0,0 +1,94 @@
...
+{
+    fn concat(&self, other: &Stream<D>) -> Stream<D>;
+}
+
+impl<D> Concat<D> for Stream<D>
+where
+    D: Data + for<'a> Deserialize<'a>,
+{
+    fn concat(&self, other: &Stream<D>) -> Stream<D> {
+        let name = format!("concat-{}-{}", self.name(), other.name());
```
sukritkalra, line 85:
```suggestion
        let name = format!("concat-{}-{}", self.id(), other.id());
```

It seems like `name()` messes up when we have a cascade of LINQ operators, so we decided to use the `id` of the stream instead.

sukritkalra: Can you also add an example `Concat` operator to the results of the `SplitOp` that we have in `linq.rs` once we pull in #193.

sukritkalra:
Will fix ROS CI issues in a separate PR. Pulling this in for now.

PR #197, opened 2022-01-26
[BUG] Fix pyo3 build issue in MacOS (#195)

sukritkalra:
Fixes #195 on x86-64 Mac. Run with `cargo build --no-default-features`. 

@objorkman: Can you test on the M1?

PR #198, opened 2022-01-26
Fix build issuse for M1 Macs

objorkman:


sukritkalra:
Closing because of same reason as #197.

PR #201, opened 2022-01-28
[LINQ] Replace MapOperator with FlatMapOperator, and add flat_map to API.

pschafhalter:
- Replaces `MapOperator` with a more general `FlatMapOperator`.
- Updates documentation.
- Fixes examples and tests.

### DIFF
--- a/erdos/src/dataflow/operators/map.rs
+++ b/erdos/src/dataflow/operators/map.rs
@@ -140,7 +140,7 @@ where
         F: 'static + Fn(&D1) -> I + Send + Sync + Clone,
         I: 'static + IntoIterator<Item = D2>,
     {
-        let op_name = format!("FlatMapOp_{}", self.name());
+        let op_name = format!("FlatMapOp_{}", self.id());
 
         crate::connect_one_in_one_out(
             move || -> FlatMapOperator<D1, _> { FlatMapOperator::new(flat_map_fn.clone()) },
--- a/erdos/src/dataflow/operators/map.rs
+++ b/erdos/src/dataflow/operators/map.rs
@@ -10,77 +10,90 @@ use crate::dataflow::{
     Data,
 };
 
-/// Maps an incoming stream of type D1 to a stream of type D2 using the provided
+/// Maps an incoming stream of type D to a stream of type `I::Item` using the provided
 /// function.
 ///
 /// # Example
-/// The below example shows how to use a MapOperator to double an incoming stream of usize
+/// The below example shows how to use a [`FlatMapOperator`] to double an incoming stream of usize
 /// messages, and return them.
 ///
-/// ```
-/// # use erdos::dataflow::{stream::IngestStream, operator::{OperatorConfig}, operators::{MapOperator}};
+/// ```F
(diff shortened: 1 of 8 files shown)


erdos/src/dataflow/operators/map.rs
```diff
@@ -10,77 +10,110 @@ use crate::dataflow::{
...
-            .unwrap();
-        tracing::debug!(
-            "{} @ {:?}: received {:?} and sent {:?}",
-            ctx.get_operator_config().get_name(),
-            ctx.get_timestamp(),
-            data,
-            (self.map_function)(data)
-        );
+    fn on_data(&mut self, ctx: &mut OneInOneOutContext<(), I::Item>, data: &D) {
+        tracing::info!("on data");
```
sukritkalra, line 64: Any particular reason this was kept separately and on `info!(...)`?

erdos/src/dataflow/operators/map.rs
```diff
@@ -10,77 +10,110 @@ use crate::dataflow::{
...
+            let timestamp = ctx.get_timestamp().clone();
+            let msg = Message::new_message(timestamp, item);
+            ctx.get_write_stream().send(msg).unwrap();
+        }
     }
 
-    fn on_watermark(&mut self, _ctx: &mut OneInOneOutContext<(), D2>) {}
+    fn on_watermark(&mut self, _ctx: &mut OneInOneOutContext<(), I::Item>) {
+        tracing::info!("{}: got watermark", _ctx.get_operator_config().get_name());
+    }
```
sukritkalra, line 82:
```suggestion
    fn on_watermark(&mut self, ctx: &mut OneInOneOutContext<(), I::Item>) {
        tracing::info!("{}: got watermark", ctx.get_operator_config().get_name());
    }
```

erdos/src/dataflow/operators/map.rs
```diff
@@ -95,7 +128,25 @@ where
...
+            self,
+        )
+    }
+
+    fn flat_map<F, I>(&self, flat_map_fn: F) -> Stream<D2>
+    where
+        F: 'static + Fn(&D1) -> I + Send + Sync + Clone,
+        I: 'static + IntoIterator<Item = D2>,
+    {
+        let op_name = format!("FlatMapOp_{}", self.name());
```
sukritkalra, line 146:
```suggestion
        let op_name = format!("FlatMapOp_{}", self.id());
```

sukritkalra (changes requested):

erdos/src/dataflow/operators/map.rs
pschafhalter: Sorry, those are debug statements I forgot to remove.

erdos/src/dataflow/operators/map.rs
pschafhalter: Why do we prefer to use the stream ID over the stream name?

sukritkalra: Same problem as before, if you cascade operators via LINQ the name ends up being `MapOp_SplitOp_FilterOp_writestream_writestream_writestream`.

sukritkalra (approved):
LGTM, with the `id()` change.

PR #202, opened 2022-01-31
[Python] Add LINQ API to streams

sguduguntla:


python/examples/linq.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def main():
+    """Creates and runs the dataflow graph."""
+    count_stream = erdos.connect_source(SendOp,
+                                        erdos.operator.OperatorConfig())
+
+    map_stream = count_stream.map(lambda a: a * 2)
+    filtered_stream = map_stream.filter(lambda a: a > 10)
+    left_stream, right_stream = filtered_stream.split(lambda a: a < 50)
```
pschafhalter, line 33: Use `x` instead of `a` for the arguments.

python/src/py_stream/py_stream.rs
```diff
@@ -22,6 +26,86 @@ impl PyStream {
...
+                    .extract()
+                    .unwrap();
+                let result: PyObject = new_fn.call1(py, (des_data,)).unwrap().extract(py).unwrap();
+                pickle
+                    .getattr("dumps")
+                    .unwrap()
+                    .call1((result,))
+                    .unwrap()
+                    .extract()
+                    .unwrap()
```
pschafhalter, line 50: It might be cleaner to do serialization/deserialization in a Python wrapper function instead of in Rust. I realize we're doing the same thing in the Python operators, so fine with keeping this as is for this PR, but we may want to make a follow-up PR to clean up serialization/deserialization.
@sukritkalra what do you think?

pschafhalter (commented):
LGTM at a high level.

Because ERDOS currently lacks scheduling, this will cause GIL conflicts because every Python LINQ operator will be placed on node 0 by default. I'm ok with merging the PR as-is and making that fix once we support resources.

sukritkalra:
Closing because #223 subsumes this change.

PR #203, opened 2022-02-01
Implement TimeVersionedState and clean up State API.

sukritkalra:
TODOS:
- Add documentation once the PR is approved.


erdos/src/dataflow/connect.rs
```diff
@@ -16,14 +16,12 @@ use crate::{
...
 /// graph by interacting with external data sources (e.g., other systems, sensor data).
-pub fn connect_source<O, S, T>(
+pub fn connect_source<O, T>(
     operator_fn: impl Fn() -> O + Clone + Send + Sync + 'static,
-    state_fn: impl Fn() -> S + Clone + Send + Sync + 'static,
     mut config: OperatorConfig,
 ) -> Stream<T>
 where
-    O: 'static + Source<S, T>,
-    S: State,
```
pschafhalter, line 26: Are there cases where we'd want to provide a system-managed state for sources? I'm thinking in case of system-managed checkpointing for fault tolerance, or operator migration.

In any case, I'm happy to make this change, and potentially re-introduce state for other purposes in a future release.

erdos/src/dataflow/context.rs
```diff
@@ -144,9 +144,27 @@ where
         &self.config
     }
 
-    /// Get the state attached to the operator.
-    pub fn get_state(&mut self) -> &mut S {
-        &mut self.state
+    /// Get the current state attached to the operator.
+    pub fn get_current_state(&mut self) -> Option<&mut S::Item> {
```
pschafhalter, line 148: I'd like to move away from `get_*` naming, and use `current_state()` instead.

Could you either:
1. Rename the functions, e.g. `get_current_state` -> `current_state`?
2. Stick with the current naming (to keep this PR focused), and make a follow-up PR changing the naming from `get_*()` to `*()`?

erdos/src/dataflow/state.rs
```diff
@@ -1,38 +1,81 @@
...
-
-// TODO: keep around messages. Add an iterator over messages.
-// Add set_timestamp and set_access_context to State.
-
-/// Trait that must be implemented by stream state.
-pub trait State: 'static + Clone + Send + Sync {}
-impl<T: 'static + Clone + Send + Sync> State for T {}
+use std::collections::HashMap;
 
 // TODO (Sukrit): Do these state traits also require a way to read the state for a given timestamp?
```
pschafhalter, line 6: Remove the TODO?

erdos/src/dataflow/state.rs
```diff
@@ -1,38 +1,81 @@
...
-pub trait StateT: 'static + Send + Sync {
+pub trait State: 'static + Send + Sync {
+    type Item: Default;
+
     fn commit(&mut self, timestamp: &Timestamp);
 
-    fn get_last_committed_timestamp(&self) -> Timestamp;
+    fn last_committed_timestamp(&self) -> Timestamp;
+
+    fn at(&mut self, timestamp: &Timestamp) -> Option<&mut Self::Item>;
```
pschafhalter, line 18: Reminder to add documentation, as you mentioned in the PR :)

erdos/src/dataflow/state.rs
```diff
@@ -1,38 +1,81 @@
...
+
+    fn at(&mut self, timestamp: &Timestamp) -> Option<&mut Self::Item> {
+        Some(self.state.entry(timestamp.clone()).or_default())
+    }
 }
 
 /// Trait that must be implemented by a state structure that is used in a Sequential operator.
 /// This state structure must implement an `append` method that enables message callbacks to add
 /// intermediate state to the structure, and a `commit` method that commits the final state for a
 /// given timestamp t.
```
pschafhalter, line 77: Could you move parts of this documentation to the methods?

Also, can you change `get_last_committed_timestamp` to `last_committed_timestamp`?

examples/full_pipeline.rs
```diff
@@ -49,39 +49,21 @@ impl SquareOperator {
...
-    }
-
-    fn get_last_committed_timestamp(&self) -> Timestamp {
-        self.current_timestamp.clone()
-    }
-}
-
-impl OneInOneOut<SquareOperatorState, usize, usize> for SquareOperator {
-    fn setup(&mut self, ctx: &mut SetupContext<SquareOperatorState>) {
+impl OneInOneOut<TimeVersionedState<()>, usize, usize> for SquareOperator {
```
pschafhalter, line 52: Why is this time-versioning a tuple? Wouldn't it make more sense to time-version a number, or provide `()` as the state?

examples/full_pipeline.rs
```diff
@@ -168,52 +135,26 @@ impl SinkOperator {
...
-        tracing::info!("SinkOperator @ {:?}: Received {}", timestamp, data);
-        ctx.get_state().increment_message_count(&timestamp);
+        tracing::info!(
+            "{} @ {:?}: Received {}",
+            ctx.get_operator_config().get_name(),
+            timestamp,
+            data
+        );
+        let current_state = ctx.get_current_state().unwrap();
+        *current_state += 1;
```
pschafhalter, line 148:
```suggestion
        *ctx.get_current_state().unwrap() += 1;
```

examples/linq.rs
```diff
@@ -48,52 +49,27 @@ impl SinkOperator {
...
-    fn on_watermark(&mut self, ctx: &mut SinkContext<SinkOperatorState>) {
+    fn on_watermark(&mut self, ctx: &mut SinkContext<TimeVersionedState<usize>>) {
         let timestamp = ctx.get_timestamp().clone();
         tracing::info!(
-            "SinkOperator @ {:?}: Received {} data messages.",
+            "{} @ {:?}: Received {} data messages.",
+            ctx.get_operator_config().get_name(),
             timestamp,
-            ctx.get_state().get_message_count(&timestamp),
+            *ctx.get_current_state().unwrap(),
```
pschafhalter, line 72:
```suggestion
            ctx.get_current_state().unwrap(),
```
Dereferencing shouldn't be necessary for displaying the value.

examples/full_pipeline.rs
```diff
@@ -227,74 +168,43 @@ impl JoinSumOperator {
...
             data,
-            state
+            *current_state
         );
     }
 
-    fn on_watermark(&mut self, ctx: &mut TwoInOneOutContext<JoinSumOperatorState, usize>) {
-        let state = ctx.get_state().get_sum();
+    fn on_watermark(&mut self, ctx: &mut TwoInOneOutContext<TimeVersionedState<usize>, usize>) {
+        let state = *ctx.get_current_state().unwrap();
```
pschafhalter, line 207:
```suggestion
        let state = ctx.get_current_state().unwrap();
```

examples/full_pipeline.rs
```diff
@@ -111,53 +93,38 @@ impl SumOperator {
...
+        // Find the last committed state, add the received data to it, and save for this timestamp.
+        {
+            let past_state = ctx
+                .get_past_state(&ctx.get_last_committed_timestamp())
+                .unwrap();
+            *ctx.get_current_state().unwrap() += past_state + data;
+        }
+
+        // Send the message.
+        let current_state = *ctx.get_current_state().unwrap();
```
pschafhalter, line 116: As-is, this copies a value, so mutating the `current_state` variable will not mutate the operator's state. Keeping `current_state` as a reference, and de-referencing/copying when creating the message should make things more clear.
```suggestion
        let current_state = ctx.get_current_state().unwrap();
```

erdos/src/dataflow/connect.rs
sukritkalra: It's not immediately clear to me how to provide *automatic* checkpointing for `Source` operators since the `run` method is supposed to be a place where the developer takes control. They could still use our `TimeVersionedState` but they would have to call their own commits and stuff.

My current take is to remove it, and if we come up with a clean way to do automatic checkpointing, we can potentially re-introduce it.

examples/full_pipeline.rs
sukritkalra: This creates a problem with multiple mutable copies since we use `ctx.get_write_stream` below.

examples/full_pipeline.rs
sukritkalra: Similar problems with multiple mutable borrows if we do this.

examples/full_pipeline.rs
pschafhalter: Oh, I see. Can we call it `current_state_copy` to clarify?

erdos/src/dataflow/connect.rs
pschafhalter: Could you update the documentation in `lib.rs` to remove the state initialization?

erdos/src/dataflow/context.rs
sukritkalra: I'll push out a new PR to do the renaming of the `get` methods.

erdos/src/dataflow/state.rs
```diff
@@ -1,43 +1,92 @@
...
-pub trait AppendableStateT<S>: 'static + Clone + Send + Sync {
+pub trait AppendableState<S>: 'static + Clone + Send + Sync {
```
pschafhalter, line 84: For API consistency, would it make sense to use an associated type here too?

erdos/src/dataflow/state.rs
```diff
@@ -1,43 +1,92 @@
...
+
+    fn at(&mut self, timestamp: &Timestamp) -> Option<&mut Self::Item> {
+        Some(self.state.entry(timestamp.clone()).or_default())
+    }
 }
 
 /// Trait that must be implemented by a state structure that is used in a Sequential operator.
 /// This state structure must implement an `append` method that enables message callbacks to add
 /// intermediate state to the structure, and a `commit` method that commits the final state for a
 /// given timestamp t.
```
pschafhalter, line 83: Could you move the documentation of the methods above the method declarations?

E.g.
```rust
   /// Adds intermediate state to the structure. Invoked by a callback.
   fn append(&self, data &S);
```

erdos/src/dataflow/state.rs
```diff
@@ -1,43 +1,92 @@
...
-pub trait AppendableStateT<S>: 'static + Clone + Send + Sync {
+pub trait AppendableState<S>: 'static + Clone + Send + Sync {
     fn append(&self, data: &S);
```
pschafhalter, line 85: Does `append()` need to take a timestamp as an argument? I'm concerned about scenarios where callbacks for multiple timestamps run simultaneously. In that case, it may also need to return an error (e.g. appending to an already-committed timestamp).

erdos/src/dataflow/state.rs
```diff
@@ -1,43 +1,92 @@
...
 
     fn commit(&self, timestamp: &Timestamp);
 
-    fn get_last_committed_timestamp(&self) -> Timestamp;
+    fn last_committed_timestamp(&self) -> Timestamp;
```
pschafhalter, line 89: Should the trait also provide a method to retrieve committed state?

erdos/src/dataflow/state.rs
```diff
@@ -1,43 +1,92 @@
...
 }
 
 // /// Error thrown upon an invalid attempt to access a portion of the
```
pschafhalter, line 92: Could you delete the commented-out portions if they are no longer useful?

python/erdos/streams.py
```diff
@@ -180,22 +194,9 @@ class IngestStream(object):
...
-    @name.setter
-    def name(self, name: str):
-        self._py_ingest_stream.set_name(name)
-
-    @property
-    def id(self) -> str:
-        """ The id of the IngestStream. """
-        return uuid.UUID(self._py_ingest_stream.id())
+    def __init__(self, name: Union[str, None] = None):
+        super().__init__(PyIngestStream(name))
```
pschafhalter, line 199: Could you rebase on the latest commit? I'm a little worried by these changes since they appear to be duplicates of commits that already occured. E.g. `IngestStream` should inherit from `Stream`, but that doesn't appear to be the case in this file.

sukritkalra: Done.

erdos/src/dataflow/state.rs
```diff
@@ -1,584 +1,90 @@
 //! Structures and traits for states added to streams.
 
 use crate::dataflow::Timestamp;
+use std::collections::HashMap;
 
-// TODO: keep around messages. Add an iterator over messages.
-// Add set_timestamp and set_access_context to State.
+/// The `State` trait must be implemented by the state exposed to the operators by ERDOS.
```
pschafhalter, line 6:
```suggestion
/// The [`State`] trait must be implemented by the state exposed to the operators by ERDOS.
```

erdos/src/dataflow/state.rs
```diff
@@ -1,584 +1,90 @@
...
-/// Trait that must be implemented by a state structure that is used in a Parallel operator.
-/// This structure must implement a `commit` method that commits the final state for a given
-/// timestamp `t`.
-pub trait StateT: 'static + Send + Sync {
+    /// The `commit` method commits the final state for a given timestamp.
     fn commit(&mut self, timestamp: &Timestamp);
 
-    fn get_last_committed_timestamp(&self) -> Timestamp;
+    /// Retrieves the last committed timestamp by this state.
+    /// This method can be used in conjunction with `at` to retrieve the latest committed state.
```
pschafhalter, line 14:
```suggestion
    /// This method can be used in conjunction with [`Self::at`] to retrieve the latest committed state.
```

erdos/src/dataflow/state.rs
```diff
@@ -1,584 +1,90 @@
...
+    fn last_committed_timestamp(&self) -> Timestamp;
+
+    /// Retrieve the state at a given timestamp.
+    /// If the state for that timestamp hasn't been initialized yet, invoke the default method on
+    /// the `Item` type, and return the newly created state for that timestamp.
```
pschafhalter, line 19:
```suggestion
    /// the [`Self::Item`] type, and return the newly created state for that timestamp.
```

pschafhalter (approved):
LGTM! Could you update the doc comments so that the generated documentation adds links to the types? And update the PR description? Then feel free to merge :)

PR #205, opened 2022-02-10
Reintroduce Stream trait, add PyStream base class

pschafhalter:
- Reintroduces the `Stream` trait to clean up the connections API.
- Adds an `OperatorStream` type which is returned when connecting operators.
- Adds a `PyStream` base class which functions similarly to the Rust `Stream` trait.
- Updates the rust toolchain due to a compiler bug present in the previous nightly.

The advantage of this PR is that it clarifies the Rust API when connecting operators, and uses type checking to prevent certain erroneous patterns that were previously possible, e.g. connecting an IngestStream to a loop by doing the following:
```rust
let ingest_stream = IngestStream::new();
let loop_stream = LoopStream::new();
loop_stream.connect_loop(&Stream::from(&ingest_stream));
```


python/erdos/__init__.py
```diff
@@ -127,9 +127,9 @@ def connect_sink(
     if isinstance(read_stream, Stream):
         py_stream = read_stream._py_stream
     elif isinstance(read_stream, IngestStream):
-        py_stream = read_stream._py_ingest_stream.to_py_stream()
+        py_stream = read_stream._py_ingest_stream
     elif isinstance(read_stream, LoopStream):
-        py_stream = read_stream._py_loop_stream.to_py_stream()
+        py_stream = read_stream._py_loop_stream
```
sukritkalra, line 132: Can we consolidate them now? We don't need to convert the streams anymore so we can just save the internal Python stream classes in all kinds of stream in `read_stream._py_stream` and skip this check.

sukritkalra (commented):
Should the `connect` methods in Python now return a `Stream` or an `OperatorStream`? Why was an `OperatorStream` class not created in Python just to make sure that the API is consistent?

python/erdos/__init__.py
```diff
@@ -73,13 +73,13 @@ def connect_source(
 
     internal_stream = _internal.connect_source(op_type, config, args, kwargs,
                                                node_id)
-    return Stream(_py_stream=internal_stream)
+    return OperatorStream(internal_stream)
```
sukritkalra, line 76: The corresponding documentation also needs to be fixed.

sukritkalra (commented):
Can you please fix the Python formatting issues?

sukritkalra (approved):
LGTM. Thanks for addressing the comments!

PR #210, opened 2022-03-18
Add infrastructure for pre-commit hooks.

sukritkalra:
- Adds a `setup_hooks.sh` script to install `pre-commit` and initialize it.
- Added [`isort`](https://github.com/PyCQA/isort) to the pre-commit hooks for automatic sorting of Python imports.
- Ran `isort` on all Python files.


setup_hooks.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
ICGog, line 1: Can you please move this file to the scripts directory?

ICGog (approved):
LGTM, but please address my comment.

pschafhalter (changes requested):
I'm leaning against enabling pre-commit hooks, as I haven't seen this feature provided in major python repos.

Could we instead add isort as a [GitHub action](https://pycqa.github.io/isort/docs/configuration/github_action.html) to our code formatting checks?

sukritkalra:
I'm leaning towards enabling them since they make sure that formatting issues get sorted before a commit instead of pushing something out and having Github Actions remind me that my build is failing. (A pattern that happens to me a lot of times since I forget to run formatting scripts when developing using vim)

pschafhalter:
Wouldn't a cleaner solution be to [set up git hooks for your local environment](https://pycqa.github.io/isort/docs/configuration/git_hook.html)? For example, I have my editor configured to automatically format on save; yet I haven't pushed my editor's config to the repository.

Moreover, supporting pre-commit hooks will result in more dependencies to manage, and an additional file to maintain to ensure consistent formatting between the hooks and the CI checks.

ICGog:
I agree with Peter. IMO, it would be better to add this check as a GitHub action since we're already using them. In this way we also avoid having to do additional setup and dependency maintenance.

.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
```
pschafhalter, line 1: Should we rename this file to `setup.cfg`? This way we can extend it if we introduce additional tools.

python/examples/join_streams.py
```diff
@@ -1,14 +1,15 @@
...
 
-import erdos
 import time
 from typing import Any
 
 from erdos.context import TwoInOneOutContext
 from erdos.operator import Source, TwoInOneOut
 from erdos.streams import WriteStream
 
+import erdos
```
pschafhalter, line 11: Could you verify that the `isort` action is functioning correctly? When I run `isort --check-only --diff --profile=black .` locally, this is moved above the `from erdos.* import ...` imports.

pschafhalter (commented):
Could you also remove the `.pylintrc` file?

python/examples/join_streams.py
sukritkalra: Weird, running `isort --check-only --diff --profile black python/examples/join_streams.py` with `isort 5.10.1` after putting it above the `from erdos.* import ...` gives me this:

```python
ERROR: /Users/sukritkalra/PhD/erdos/python/examples/join_streams.py Imports are incorrectly sorted and/or formatted.
--- /Users/sukritkalra/PhD/erdos/python/examples/join_streams.py:before 2022-03-23 10:13:18.521570
+++ /Users/sukritkalra/PhD/erdos/python/examples/join_streams.py:after  2022-03-23 10:13:22.709842
@@ -4,12 +4,11 @@
 import time
 from typing import Any

-import erdos
-
 from erdos.context import TwoInOneOutContext
 from erdos.operator import Source, TwoInOneOut
 from erdos.streams import WriteStream

+import erdos


 class SendOp(Source):
```

.flake8
sukritkalra: Agreed and done.

pschafhalter (approved):
LGTM

PR #211, opened 2022-03-22
[Bug] Stream.name() -> Stream.name, enable access to global AbstractGraph from multiple threads

pschafhalter:
Fixes bugs affecting the Pylot redesign.
1. Some pieces of Python code were calling the method `stream.name()`, which has been deprecated in favor of the property `stream.name`.
2. The `AbstractGraph` was inaccessible from different threads. This causes errors, e.g. when an `IngestStream` is moved to another thread in order to send messages from a CARLA-managed callback, and the `IngestStream`'s name is accessed.


ICGog (approved):
LGTM

sukritkalra (approved):
LGTM.

PR #213, opened 2022-03-24
[CI] Enable Clippy linting and separate formatting checks.

sukritkalra:
- Enables Clippy linting on the Rust code, and fixed code to conform to the warnings raised. There seems to be a bug in the Clippy linting that does not respect an `#[allow(needless_option_as_deref)]` on the `new` method of the `PyMessage` class, so it was added as a global allow for now. 
- Separates Formatting checks for Rust and Python into separate jobs, and makes the builds depend on the formatting checks to succeed first.
- Package builds are now cached at the formatting stage, which allows the Rust/ROS/Python builds to use the cache instead of rebuilding the dependencies in parallel.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -69,7 +69,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.10
+          python-version: "3.10"
       - name: Check Python imports
         uses: isort/isort-action@master
         with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,21 +60,16 @@ jobs:
           source /opt/ros/$ROS_DISTRO/setup.bash
           cargo test --verbose --features=ros
 
-  build-python:
-    name: "Python ${{ matrix.python-version }} Build"
+  format-python:
+    name: "Python Formatting"
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
     steps:
-      - uses: actions/checkout@v2
-      - name: Setup python
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup Python
         uses: actions/setup-python@v2
         with:
-          python-version: ${{ matrix.python-version }}
-      - name: Install Python dependencies
-        run: |
-          python -m pip install maturin toml flake8==3.9.2 flake8-quotes
+          python-version: 3.10
       - name: Check Python imports
         uses: isort/isort-action@master
         with:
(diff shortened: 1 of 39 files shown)


erdos/src/node/mod.rs
```diff
@@ -11,6 +11,7 @@
 //! scheduling operators, and hope to provide a versatile solution.
 
 // Private submodules
+#[allow(clippy::module_inception)]
 mod node;
```
pschafhalter, line 15: Need to fix this in a future PR. I added a corresponding task.

.github/workflows/ci.yaml
```diff
@@ -11,21 +11,42 @@ env:
...
           override: true
-          components: rustfmt, rust-src
-      - uses: Swatinem/rust-cache@v1
+          components: rustfmt, clippy
+      - name: Cache Rust package builds
+        uses: Swatinem/rust-cache@v1
       - name: Check Rust formatting
         run: cargo fmt -- --check
+      - name: Check Clippy linting
+        run: cargo clippy --all --tests --all-features -- -A clippy::needless_option_as_deref
```
pschafhalter, line 32: Can fix the `clippy::needless_option_as_deref` issue with `PyMessage` by upgrading PyO3 to [v0.16](https://pyo3.rs/master/changelog.html#fixed-2). I added a task to address this in a future PR.

erdos/src/node/operator_event.rs
```diff
@@ -94,10 +94,7 @@ impl Eq for OperatorEvent {}
 
 impl PartialEq for OperatorEvent {
     fn eq(&self, other: &OperatorEvent) -> bool {
-        match self.cmp(other) {
-            Ordering::Equal => true,
-            _ => false,
-        }
+        matches!(self.cmp(other), Ordering::Equal)
```
pschafhalter, line 97:
```suggestion
        self.cmp(other).is_eq()
```

erdos/src/scheduler/endpoints_manager.rs
```diff
@@ -47,6 +47,12 @@ pub struct ChannelsToSenders {
     senders: HashMap<NodeId, UnboundedSender<InterProcessMessage>>,
 }
 
+impl Default for ChannelsToSenders {
```
pschafhalter, line 50: Derive Default.

python/src/py_operators/mod.rs
```diff
@@ -26,30 +26,36 @@ fn construct_operator(
 ) -> Arc<PyObject> {
     Python::with_gil(|py| -> Arc<PyObject> {
         let locals = PyDict::new(py);
-        locals
+        if let Some(e) = locals
```
pschafhalter, line 29: Could you add a TODO that this function should return a Result? This isn't a good way to handle errors.

pschafhalter (approved):
Thanks for this PR! LGTM once the comments are addressed!

PR #214, opened 2022-03-25
[CI] Fail on warnings when building Rust docs

pschafhalter:
- The rustdoc check fails if building the documentation produces any warnings.
- Enables all features when building the documentation.
- Fixes errors in the documentation, add iterates on ROS operator documentation.
- Adds caching to the rustdoc workflow, reducing run times in CI from ~2.5 minutes to ~30 seconds.


sukritkalra (approved):
LGTM

PR #215, opened 2022-03-25
[LINQ] Add timestamp join

pschafhalter:
Introduces the `TimestampJoin`, which joins messages with matching timestamps from two different streams by performing a cartesian product. I've tested the operator locally, and it seems to work.

A few things I wasn't sure about that can hopefully be addressed in review:
1. Consistent naming: `TimestampJoin` seems more ergonomic, but this should be named `TimestampJoinOperator` to stay consistent with the other operators.
2. Example: I wasn't sure how to best add the timestamp join to the existing LINQ and full pipeline examples. These examples may also need to be revamped.

### DIFF
--- a/erdos/src/dataflow/operators/join.rs
+++ b/erdos/src/dataflow/operators/join.rs
@@ -11,10 +11,10 @@ use crate::dataflow::{
 
 /// Joins messages with matching timestamps from two different streams.
 ///
-/// The following table provides an example of how the [`TimestampJoin`] processes data from two
+/// The following table provides an example of how the [`TimestampJoinOperator`] processes data from two
 /// streams:
 ///
-/// | Timestamp | Left input | Right input | [`TimestampJoin`] output                   |
+/// | Timestamp | Left input | Right input | [`TimestampJoinOperator`] output                   |
 /// |-----------|------------|-------------|--------------------------------------------|
 /// | 1         | a <br> b   | 1 <br> 2    | (a, 1) <br> (a, 2) <br> (b, 1) <br> (b, 2) |
 /// | 2         | c          |             |                                            |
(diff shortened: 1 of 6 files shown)


erdos/src/dataflow/operators/join.rs
```diff
@@ -1,198 +1,146 @@
...
+        left_items.push(data.clone());
+
+        let messages: Vec<_> = right_items
+            .iter()
+            .cloned()
+            .map(|right_item| Message::new_message(timestamp.clone(), (data.clone(), right_item)))
+            .collect();
+        for msg in messages {
+            ctx.get_write_stream().send(msg).unwrap();
+        }
```
sukritkalra, line 76: Do you think we should send messages one by one instead of cloning them all first? My concern is that the downstream operators are blocked until the cloning of all the cartesian products takes place (which for some datatypes might take a long time)

sukritkalra (changes requested):
1. I'm in favor of renaming it to `TimestampJoinOperator` to remain consistent with the rest of the operators. Maybe at some point we could have a generic `JoinOperator` whose joining behavior (i.e., window, tumbling, timestamp etc.) could be chosen through an `enum`.
2. We have a `JoinSumOperator` in the `full_pipeline` example that could be replaced with a `TimestampJoinOperator` followed by a `MapOperator`. (I believe Flink does this using an `apply` method, which we could also provide in the LINQ API and just use a `MapOperator` underneath)

erdos/src/dataflow/operators/join.rs
pschafhalter: Agreed, but unfortunately that doesn't compile naiively because a mutable borrow of the write stream (via the context) conflicts with the mutable borrow of the state (via the context).

I'll try to come up with another solution.

pschafhalter:
I revamped the LINQ example with the timestamp join.

Do we also want to include it in the full pipeline example? If so, can you let me know the purpose of the full pipeline example? Is the intent to mirror the LINQ example using the `connect_*` functions, or is it trying to demonstrate operator implementation as well? I notice there are several unused and redundant operators (e.g. `SquareOperator` could be replaced by a `FlatMapOperator`).

sukritkalra (commented):
Yeah, the intent of the `full_pipeline` example was to show the operator construction and connection API, which is simplified by the LINQ API since they are all data transformation operator. Maybe we can leave it untouched for now, since we're planning to replace that with a Perception example anyways.

Also, what is the behavior of a `TimestampJoinOperator` when you get no inputs from one of the streams for a timestamp?

pschafhalter:
Messages are dropped when there are no inputs from one of the streams for a given timestamp. The behavior is documented in the table above the `TimestampJoinOperator`.

sukritkalra (approved):
LGTM. Can you add a comment on the `connect` methods about why we require them to take a state initialization just in case we forget again?

PR #223, opened 2022-04-30
[Python] Expose ERDOS' LINQ API in Python

sukritkalra:
- Provides the `map`, `flat_map`, `filter`, `split`, `timestamp_join` and `concat` methods from Rust in Python.
- Enables developers to provide multiple streams to `concat` to ease the implementation of multi-in operators.

Potential TODOs:
- Syntactic sugar over the `split` method to enable developers to specify the type of outputs expected and automatically route them to different streams.
- Our Rust `timestamp_join` does not allow developers to specify a `join_function` similar to `apply` in Flink, which could prevent the addition of an extra `map` when doing `timestamp_join`s in Python.

### DIFF
--- a/python/erdos/streams.py
+++ b/python/erdos/streams.py
@@ -2,7 +2,7 @@ import logging
 import pickle
 import uuid
 from abc import ABC
-from typing import Union
+from typing import Any, Callable, Union
 
 from erdos.internal import (
     PyExtractStream,
@@ -57,6 +57,25 @@ class Stream(ABC):
     @name.setter
     def name(self, name: str):
         self._internal_stream.set_name(name)
+
+    def map(self, function: Callable[[Any], Any]) -> "OperatorStream":
+        """Applies the given function to each received value on the stream, and outputs
+        the results on the returned stream.
+
+        Args:
+            function (Callable[[Any], Any]): The function to be applied on each message
+                received on the input stream.
+
+        Returns:
+            An :py:class:`OperatorStream` that carries the results of the applied
+            function.
+        """
+
+        def map_fn(serialized_data: bytes) -> bytes:
+            result = function(pickle.loads(serialized_data))
+            return pickle.dumps(result)
+
+        return OperatorStream(self._internal_stream._map(map_fn))
 
 
 class ReadStream:
--- a/python/erdos/streams.py
+++ b/python/erdos/streams.py
@@ -148,7 +148,7 @@ class Stream(ABC):
                 self.
 
         Returns:
-            A :py:class:`OperatorStream` that carries the joined results from the two
+            An :py:class:`OperatorStream` that carries the joined results from the two
             streams.
         """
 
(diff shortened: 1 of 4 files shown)


python/src/py_stream/mod.rs
```diff
@@ -36,6 +39,94 @@ impl PyStream {
     fn id(&self) -> String {
         format!("{}", self.id)
     }
+
+    fn _map(&self, py: Python<'_>, function: PyObject) -> PyResult<Py<PyOperatorStream>> {
```
pschafhalter, line 43:
```suggestion
    fn map(&self, py: Python<'_>, function: PyObject) -> PyResult<Py<PyOperatorStream>> {
```

`PyStream` is an internal API already, so we don't need to indicate its methods as internal.

python/erdos/streams.py
```diff
@@ -58,6 +59,147 @@ def name(self) -> str:
...
+            function.
+        """
+
+        def map_fn(serialized_data: bytes) -> bytes:
+            result = function(pickle.loads(serialized_data))
+            return pickle.dumps(result)
+
+        return OperatorStream(self._internal_stream._map(map_fn))
+
+    def flat_map(self, function: Callable[[Any], Sequence[Any]]) -> "OperatorStream":
```
pschafhalter, line 81: Could you add a todo to optimize the Python flat map implementation using iterators? In this implementation, all new elements are generated and then sent; whereas the Rust implementation sends messages eagerly after each element is generated.

python/erdos/streams.py
```diff
@@ -58,6 +59,147 @@ def name(self) -> str:
...
+    def split(
+        self, function: Callable[[Any], bool]
+    ) -> Tuple["OperatorStream", "OperatorStream"]:
+        """Applies the given function to each received value on the stream, and outputs
+        the value to either the left or the right stream depending on the returned
+        boolean value from the function.
+
+        Args:
+            function (Callable[[Any], bool]): The function to be applied to each
+                message received on the input stream.
```
pschafhalter, line 130: Could you note which stream the boolean values correspond to? It's currently not clear if `True` corresponds to the left stream or the right stream.

python/erdos/streams.py
```diff
@@ -58,6 +59,147 @@ def name(self) -> str:
...
+        def join_fn(serialized_data_left: bytes, serialized_data_right: bytes) -> bytes:
+            left_data = pickle.loads(serialized_data_left)
+            right_data = pickle.loads(serialized_data_right)
+            return pickle.dumps((left_data, right_data))
+
+        return OperatorStream(
+            self._internal_stream._timestamp_join(other._internal_stream, join_fn)
+        )
+
+    def concat(self, other: Union["Stream", Sequence["Stream"]]) -> "OperatorStream":
```
pschafhalter, line 164:
```suggestion
    def concat(self, *other: "Stream") -> "OperatorStream":
```
IMO this is a good application of variadics.

pschafhalter (changes requested):

python/src/py_stream/mod.rs
sukritkalra: I used `_map` because otherwise there is a name clash from the `map` method being implemented atop the `Stream` trait in Rust.

When creating a new `PyOperatorStream` below, the method then defaults to this `map` implementation instead of the trait one.

pschafhalter (approved):
LGTM!

PR #227, opened 2022-05-08
[Python] Update stream documentation

pschafhalter:
Don't need to specify arg and return types in docstrings as these are inferred from the type annotations.


python/out.gv
```diff
@@ -0,0 +1,21 @@
+digraph erdos_dataflow {
```
ICGog, line 1: Was this file included by mistake or are you planning on using it in an example?

pschafhalter: Added by mistake. Removed it in a follow-up commit.

PR #229, opened 2022-06-02
[Bug] Fix watermark logic

pschafhalter:
- Throw error when sending messages with timestamps equal to the low watermark.
- Fix an issue for advancing the minimum watermark in 2 in 1 out operators.
- Fix an issue where stream IDs were used instead of stream names.
- Fix an issue where data instead of a message was returned when reading from streams in Python. 

cc @objorkman


sukritkalra (approved):
LGTM

PR #230, opened 2022-06-08
[Bug] Fix deadlock in watermark callbacks

objorkman:
Fix an issue when `flow_watermarks=False` does not execute watermark callbacks.


pschafhalter (approved):
LGTM

PR #233, opened 2022-06-22
[WIP] Convert AbstractGraph to GraphBuilder

objorkman:


erdos/src/dataflow/graph/graph.rs
```diff
@@ -0,0 +1,342 @@
+use std::sync::{Arc, Mutex};
```
pschafhalter, line 1: Can we move this file into `graph/mod.rs`? I know we do this in `node.rs`, but it's [bad style](https://rust-lang.github.io/rust-clippy/master/#module_inception).

erdos/src/dataflow/graph/graph.rs
```diff
@@ -0,0 +1,342 @@
...
+}
+
+impl Graph {
+    pub fn new() -> Self {
+        Self {
+            internal_graph: Arc::new(Mutex::new(InternalGraph::new())),
+        }
+    }
+
+    pub fn get_ingest_stream<D>(&self, name: &str) -> IngestStream<D>
```
pschafhalter, line 36: I'm not a fan of these names. I'd prefer `add_ingest_stream()` or even `ingest_stream()`, since this function creates rather than retrieves. We may even want to go further with renaming, e.g. `add_ingress()` and `add_egress()` since this is more in-line with graph terminology, and leaves the door open to nested graphs.

@sukritkalra please let me know your thoughts as well!

erdos/src/dataflow/graph/graph.rs
```diff
@@ -0,0 +1,342 @@
...
+        let ingest_stream = IngestStream::new(name, Arc::clone(&self.internal_graph));
+        self.internal_graph
+            .lock()
+            .unwrap()
+            .add_ingest_stream(&ingest_stream);
+
+        ingest_stream
+    }
+
+    pub fn get_extract_stream<D>(&self, name: &str, stream: &OperatorStream<D>) -> ExtractStream<D>
```
pschafhalter, line 49: API-wise, it might make more sense to add a `to_extract_stream()` function to the `OperatorStream` struct, and make this function `pub(crate)`.

Also, I don't think it makes sense to give the extract stream a separate name, as it's like a `ReadStream` that happens to live in the driver, rather than a whole different stream that originates from an operator.

erdos/src/dataflow/stream/extract_stream.rs
```diff
@@ -159,7 +142,7 @@ where
     }
 
     pub fn name(&self) -> String {
-        default_graph::get_stream_name(&self.id)
+        self.name.clone()
```
pschafhalter, line 145: If, as per the previous comment, extract streams should inherit their parent stream name, then we need to retrieve the name from the graph to avoid a consistency issue. Consider the following:
```rust
stream.set_name("s1");
let extract_stream = stream.to_extract_stream(); // Initializes the extract stream with name s1.
stream.set_name("s2");
println!("stream name: {}" stream.name()); // s2
println!("extract_stream name: {}", extract_stream.name()); // s1
```

erdos/src/dataflow/stream/loop_stream.rs
```diff
@@ -26,34 +29,27 @@ where
     for<'a> D: Data + Deserialize<'a>,
 {
     id: StreamId,
+    name: String,
```
pschafhalter, line 32: Loop streams also shouldn't have a separate name. They should inherit the name from whichever stream they're connected to when the loop is closed. Otherwise the following presents a consistency issue:
```rust
let loop_stream = graph.add_loop_stream("s1");`
let stream = loop_stream.map(|x| 2 * x);
stream.set_name("s2");
loop_stream.connect(&stream);
```

pschafhalter (changes requested):
Adding name fields to the stream structs (`OperatorStream`, `ExtractStream`, etc) will cause inconsistencies with stream names. For example, a `LoopStream`, `OperatorStream`, and `ExtractStream` might all have different names for the same underlying stream.

I think that as before, the `InternalGraph` needs to act as a central source of truth on stream metadata (such as names), at least until the graph is compiled.

erdos/src/dataflow/graph/graph.rs
sukritkalra: I pushed an update to the visibility of the modules, and I agree with the bad styling, but don't necessarily agree that it should be moved to `mod.rs`. The style guide just says that imports shouldn't be of the form `use graph::graph::Graph`, but we could have `pub(crate) graph` in `graph/mod.rs` and publicly export `Graph` from inside the second graph module. 

I feel like the graph module's `mod.rs` is being used for abstract operator and stream types and corresponding traits, which don't necessarily go well with placing the publicly-visible `Graph` abstraction with them.

Tokio seems to be doing something similar, although not exactly the same with their `stream_ext` module [here](https://github.com/tokio-rs/tokio/blob/master/tokio-stream/src/lib.rs).

erdos/src/dataflow/graph/graph.rs
sukritkalra: I would like to move it to `add_ingress`, but I'm also in favor of removing `*_extract_stream` methods from the `graph`, and just adding a LINQ style `extract` method on `OperatorStream`s instead. However, `add_ingress` and `extract` don't really seem to go well together.

Can we change this to `add_ingest_stream()` for now, and revisit this when we build a `SubGraph` abstraction? (which should hopefully be soon)

erdos/src/dataflow/graph/graph.rs
sukritkalra: Agree with the first, we should probably remove `get_extract_stream`, and rename it to `extract`, which then also allows us to provide utility functions like `extract_to_file` to log stream contents to the disk etc.

erdos/src/dataflow/stream/extract_stream.rs
sukritkalra: We do indeed have a bigger problem here with the naming. Just as a reference, this PR removes the `set_name` method, because even apart from `ExtractStream`s, the usage of the graph to hold a central point of truth is kind of confusing. Consider, a slightly different example than the one you posted, which would occur if we let developers set names:

```rust
let source_stream = graph.connect_source(...);
source_stream.set_name("MapSourceStream");
let map_stream = graph.connect_one_in_one_out(..., &source_stream);
source_stream.set_name("FilterSourceStream");
let filter_stream = graph.connect_one_in_one_out(..., &source_stream);
````

I'm assuming that in this scenario, if we update names in the graph, that the logs would show that the `MapOperator` is receiving messages on a stream titled `FilterSourceStream`. Is my understanding correct?

File: erdos/src/dataflow/stream/loop_stream.rs (hunk 5 above)
Comment (sukritkalra) [reply erdos/src/dataflow/stream/loop_stream.rs:32]: I agree, we should probably not name the `LoopStream` since they are just virtual streams that will be removed after the graph's `compile` method is called.

File: erdos/src/dataflow/graph/graph.rs (hunk 2 above)
Comment (pschafhalter) [reply erdos/src/dataflow/graph/graph.rs]: What about `add_ingress` and `add_egress`? And then adding `to_egress()` on streams. We might also want to rename `IngestStream` to `IngressStream` and `ExtractStream` to `EgressStream`.

File: erdos/src/dataflow/stream/extract_stream.rs (hunk 4 above)
Comment (pschafhalter) [reply erdos/src/dataflow/stream/extract_stream.rs:144]: Yes, that understanding is correct, but in my mind it makes sense. I think there are a few potential solutions:
1. Print a warning (or panic) the 2nd time a stream's name is set.
2. `set_name` returns a `NamedOperatorStream`, which doesn't allow the name to be changed again.

It doesn't look like Flink supports naming streams. ROS has naming built-in with topics.

Comment (sukritkalra) [reply erdos/src/dataflow/stream/extract_stream.rs:144]: I am more in favor of removing stream names completely, and giving them names according to the operator they are arising out of. `NamedOperatorStream` seems like an abstraction that will just mirror `OperatorStream` but with one method removed.

File: erdos/src/dataflow/graph/graph.rs (hunk 2 above)
Comment (sukritkalra) [reply erdos/src/dataflow/graph/graph.rs]: I agree with this change. @objorkman could you make the corresponding changes please?

File: erdos/src/dataflow/stream/extract_stream.rs (hunk 4 above)
Comment (pschafhalter) [reply erdos/src/dataflow/stream/extract_stream.rs:144]: Hmm, I think this can still make debugging difficult. Consider some of the patterns in Pylot, where an operator releases a Union of several types, and these are then split by type. Giving these streams sensible names (rather than something based on the output operator) is very useful for debugging.

File: erdos/src/dataflow/stream/loop_stream.rs (hunk 5 above)
Comment (pschafhalter) [reply erdos/src/dataflow/stream/loop_stream.rs]: Counterpoint: it is useful to name loops for debugging (e.g. when a loop is unconnected). Also, if we apply LINQ operations to a loop stream, we might have trouble naming the LINQ operators.

File: erdos/src/dataflow/operators/filter.rs [hunk 6]
```diff
@@ -93,11 +94,17 @@ where
     {
         let op_name = format!("FilterOp_{}", self.id());
 
-        crate::connect_one_in_one_out(
+        let write_stream = OperatorStream::new(
+            &format!("{}-write-stream", op_name),
+            Arc::clone(&self.get_graph()),
+        );
```
sukritkalra, line 100: Extremely minor OCD nit: In the rest of the LINQ operators, we have an empty line after `let write_stream` instead of before.

erdos/src/dataflow/operators/map.rs
```diff
@@ -125,15 +129,21 @@ where
     {
         let op_name = format!("MapOp_{}", self.id());
 
-        crate::connect_one_in_one_out(
+        let write_stream = OperatorStream::new(
+            &format!("{}-write-stream", op_name),
+            Arc::clone(&self.get_graph()),
+        );
```
sukritkalra, line 135: Similar to the extremely minor OCD nit before.

erdos/src/dataflow/stream/egress_stream.rs
```diff
@@ -6,117 +6,102 @@ use std::{
...
 where
     for<'a> D: Data + Deserialize<'a>,
 {
-    /// Returns a new instance of the [`ExtractStream`].
+    /// Returns a new instance of the [`EgressStream`].
     ///
     /// # Arguments
     /// * `stream`: The [`Stream`] returned by an [operator](crate::dataflow::operator)
     /// from which to extract messages.
     pub fn new(stream: &OperatorStream<D>) -> Self {
```
sukritkalra, line 86:
```suggestion
    pub(crate) fn new(stream: &OperatorStream<D>) -> Self {
```

erdos/src/dataflow/stream/ingress_stream.rs
```diff
@@ -0,0 +1,159 @@
+use std::{
+    sync::{Arc, Mutex},
+    thread,
+    time::Duration,
+};
+
```
sukritkalra, line 6: Can you git rename this file instead of deleting and restarting it? It would be nice to have commit history.

erdos/src/dataflow/stream/ingress_stream.rs
```diff
@@ -0,0 +1,159 @@
...
+where
+    for<'a> D: Data + Deserialize<'a>,
+{
+    // The unique ID of the stream (automatically generated by the constructor)
+    id: StreamId,
+    // The name of the stream
+    name: String,
+    // Use a std mutex because the driver doesn't run on the tokio runtime.
+    write_stream_option: Arc<Mutex<Option<WriteStream<D>>>>,
+    //
```
sukritkalra, line 72: Please add a comment.

erdos/src/dataflow/stream/ingress_stream.rs
```diff
@@ -0,0 +1,159 @@
...
+            .as_ref()
+            .map(WriteStream::is_closed)
+            .unwrap_or(true)
+    }
+
+    /// Sends data on the stream.
+    ///
+    /// # Arguments
+    /// * `msg` - The message to be sent on the stream.
+    pub(crate) fn send(&mut self, msg: Message<D>) -> Result<(), SendError> {
```
sukritkalra, line 114: Does this work with `pub(crate)`?

erdos/src/dataflow/stream/ingress_stream.rs
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    pub(crate) fn get_write_stream(&self) -> Arc<Mutex<Option<WriteStream<D>>>> {
+        Arc::clone(&self.write_stream_option)
+    }
+
+    /// Returns `true` if a top watermark message was received or the [`IngressStream`] failed to
+    /// set up.
+    pub(crate) fn is_closed(&self) -> bool {
```
sukritkalra, line 101: We probably should be exposing this to developers, similar to `EgressStream`.

erdos/src/dataflow/stream/loop_stream.rs
```diff
@@ -27,41 +31,39 @@ where
...
+        );
     }
-}
 
-impl<D> Stream<D> for LoopStream<D>
-where
-    for<'a> D: Data + Deserialize<'a>,
-{
-    fn id(&self) -> StreamId {
+    pub(crate) fn get_id(&self) -> StreamId {
```
sukritkalra, line 66: What convention are we using in the API? I seem to remember everything is `id()` instead of `get_id()`

erdos/src/dataflow/stream/mod.rs
```diff
@@ -55,37 +59,48 @@ pub trait WriteStreamT<D: Data> {
...
         Self {
-            id,
+            id: StreamId::new_deterministic(),
+            name: name.to_string(),
             phantom: PhantomData,
+            graph,
         }
     }
+
+    pub(crate) fn to_egress(&self) -> EgressStream<D>
```
sukritkalra, line 88:
```suggestion
    pub fn to_egress(&self) -> EgressStream<D>
```

erdos/src/dataflow/stream/mod.rs
```diff
@@ -55,37 +59,48 @@ pub trait WriteStreamT<D: Data> {
...
 pub trait Stream<D: Data> {
-    fn name(&self) -> String {
-        default_graph::get_stream_name(&self.id())
-    }
-    fn set_name(&mut self, name: &str) {
-        default_graph::set_stream_name(&self.id(), name);
-    }
+    fn name(&self) -> String;
     fn id(&self) -> StreamId;
+    fn get_graph(&self) -> Arc<Mutex<InternalGraph>>;
```
sukritkalra, line 64:
```suggestion
    fn graph(&self) -> Arc<Mutex<InternalGraph>>;
```

Since everything seems to be following the convention of not having `get_` in the prefix, can you change this too?

erdos/src/lib.rs
```diff
@@ -209,13 +210,12 @@ impl fmt::Display for Uuid {
...
 
-/// Resets seed and creates a new dataflow graph.
+/// Resets seed.
 pub fn reset() {
     // All global variables should be reset here.
     RNG.with(|rng| {
         *rng.borrow_mut() = StdRng::from_seed(&[1913, 3, 26]);
     });
-    dataflow::graph::default_graph::set(dataflow::graph::AbstractGraph::new());
 }
```
sukritkalra, line 219: We don't need this function anymore, this was being used to reset the default graph.

examples/full_pipeline.rs
```diff
@@ -52,15 +52,15 @@ impl SquareOperator {
...
+            move |_s: &(), _t: &Timestamp| -> Duration { Duration::new(1, 0) },
             |_s: &(), _t: &Timestamp| {
                 tracing::info!("SquareOperator @ {:?}: Missed deadline.", _t);
             },
         ));
     }
 
     fn on_data(&mut self, ctx: &mut OneInOneOutContext<(), usize>, data: &usize) {
-        thread::sleep(Duration::new(2, 0));
+        thread::sleep(Duration::new(1, 0));
```
sukritkalra, line 63: Any particular reason for these changes?

sukritkalra (changes requested):

erdos/src/dataflow/stream/loop_stream.rs
objorkman: Switching to id() as the style of `get_` is not used elsewhere

erdos/src/dataflow/stream/ingress_stream.rs
objorkman: Good catch, need to be `pub`

erdos/src/dataflow/graph/internal_graph.rs
```diff
@@ -0,0 +1,745 @@
...
+        ParallelSinkMessageProcessor, ParallelTwoInOneOutMessageProcessor, SinkMessageProcessor,
+        SourceExecutor, TwoInExecutor, TwoInOneOutMessageProcessor,
+    },
+    scheduler::channel_manager::ChannelManager,
+    OperatorConfig, OperatorId,
+};
+
+use super::{
+    job_graph::JobGraph, AbstractOperator, AbstractStream, AbstractStreamT, StreamSetupHook,
+};
```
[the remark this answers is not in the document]
pschafhalter: Let's use `crate` since it groups together all ERDOS-related imports.

sukritkalra (approved):
LGTM. Thanks for the changes!

examples/full_pipeline.rs
```diff
@@ -247,86 +247,92 @@ fn main() {
...
+
+    // let map_config = OperatorConfig::new().name("FlatMapOperator");
+    // let map_stream = graph.connect_one_in_one_out(
+    //     || -> FlatMapOperator<usize, _> {
+    //         FlatMapOperator::new(|x: &usize| std::iter::once(2 * x))
+    //     },
+    //     || {},
+    //     map_config,
+    //     &square_stream,
+    // );
```
pschafhalter, line 267: Why are these commented out?

objorkman: Not sure, I'll uncomment them as they are working

